PR #29, opened 2020-09-18
feat: update management canister interface

chenyan-dfinity:


krpeacock:
@chenyan-dfinity This PR has been chilling here since October. Does this still need to happen?

chenyan-dfinity:
It's probably out-dated, I will close it. But it would be nice to update the management canister interface. We only have a small subset of the real ic interface in agent-js.

Issue #222, opened 2021-02-17
Remove dependencies from third party libraries


hansl:
A naïve webpack configuration will show that packages we use but don't bundle with needs to have a resolver fallback configuration, which is not very intuitive and impose configuration rules to applications. Instead, we should try to either bundle those libraries or (preferably) remove the dependencies if they're small and easy to get rid of.

Example:

![image](https://user-images.githubusercontent.com/681969/108283794-cc9ed280-7138-11eb-9b8c-df5d5bc96202.png)

PR #250, opened 2021-02-25
chore: remove buffer-pipe as a dependency

hansl:
And re-implement it internally. This is a very trivial piece of code that
is a step towards getting rid of Buffer itself through the codebase.

Reference #222.


krpeacock:
No longer relevant with the 0.10.0 refactor

PR #439, opened 2021-06-03
fix: Fix an IDL bug in using FixedNat with BigInt

ninegua:
The bug was `IDL.Nat64.covariant(0n) == false`, but it should be true.


chenyan-dfinity (approved):

hansl (approved):

krpeacock (approved):

ninegua:
bors merge

PR #443, opened 2021-06-03
chore: uses public ic-ref and removes secret

krpeacock:


nomeata:
The Haskell files moved up one directory, no longer in `impl/` but just in `/`

hansl (approved):

PR #451, opened 2021-06-08
test: httpagent getDefaultFetch test cases

krpeacock:
Increasing unit test coverage
67.03% -> 68.16%


hansl (approved):
LGTM. Can we automate reporting of code coverage? It's not a very useful metrics but it's motivating to see progress!

krpeacock:
Yeah, I'd personally prefer to not have it be a blocking metric, but it's useful to get reminded if your code is adding or detracting from the testing effort

hansl:
Right. To be clear I don’t want to block. But a comment about the coverage before and after could be great. Motivates people to write tests. Everyone loves numbers going up :D

PR #457, opened 2021-06-10
chore: reducing use of Buffer

krpeacock:
Also creates some utility methods for working with Uint8Arrays and ArrayBuffers

### DIFF
--- a/packages/agent/src/certificate.ts
+++ b/packages/agent/src/certificate.ts
@@ -2,16 +2,9 @@ import { Buffer } from 'buffer/';
 import { Agent, getDefaultAgent, ReadStateResponse } from './agent';
 import * as cbor from './cbor';
 import { AgentError } from './errors';
-import { hash } from './request_id';
-import {
-  BinaryBlob,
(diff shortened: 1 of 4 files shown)


packages/agent/src/certificate.ts
```diff
@@ -161,72 +142,69 @@ export class Certificate {
     if (!lookup) {
       throw new Error(`Could not find subnet key for subnet 0x${d.subnet_id.toString('hex')}`);
     }
-    return Buffer.from(lookup);
+    return lookup;
   }
 }
 
-const DER_PREFIX = Buffer.from(
+const DER_PREFIX = hexToUint8(
```
codelemur, line 149: I'm a little confused here, it used to be Buffer.from('blah','hex') so the result would be a Buffer. 

But now this is instead UInt8Array, and that works fine?

krpeacock: I also changed the one place that referenced it

codelemur (approved):
OK, don't have any additional feedback, this looks good.

PR #460, opened 2021-06-11
chore: excludes tests from tsconfig build process

krpeacock:
also allows you to do unallowed things like modifying window for tests


codelemur (approved):
Thanks

PR #463, opened 2021-06-28
fix: Implements DER wrapping and unwrapping

kritzcreek:
This is needed to be able to use the WebAuthn identity with public keys larger than 127 bytes. It should be a non-breaking change, unless someone relied on identity creation failing for larger keys.

So far this library hardcoded the DER prefixes, which works until the payload becomes larger than 127 bytes, as you need to encode the length of sequences and bitstrings according to the "Definite scheme" as described here: https://en.wikipedia.org/wiki/X.690#Definite_form

Along the same line we need to unwrap DER in other places, which runs into the same issue. So far this library pretends we can always slice off a fixed amount of bytes at the front to remove the DER wrapper, but we actually need to parse just enough structure to know how many bytes are used to encode lengths.

### DIFF
--- a/packages/identity/src/identity/der.ts
+++ b/packages/identity/src/identity/der.ts
@@ -55,22 +55,35 @@ const decodeLenBytes = (buf: Uint8Array, offset: number): number => {
   throw new Error('Length too long (> 4 bytes)');
 };
 
+/**
+ * A DER encoded `SEQUENCE(OID)` for DER-encoded-COSE
+ */
 export const DER_COSE_OID = Uint8Array.from([
   ...[0x30, 0x0c], // SEQUENCE
   ...[0x06, 0x0a], // OID with 10 bytes
   ...[0x2b, 0x06, 0x01, 0x04, 0x01, 0x83, 0xb8, 0x43, 0x01, 0x01], // DER encoded COSE
 ]);
 
+/**
+ * A DER encoded `SEQUENCE(OID)` for the Ed25519 algorithm
+ */
 export const ED25519_OID = Uint8Array.from([
   ...[0x30, 0x05], // SEQUENCE
   ...[0x06, 0x03], // OID with 3 bytes
   ...[0x2b, 0x65, 0x70], // id-Ed25519 OID
 ]);
 
-export const wrapDER = (cose: ArrayBuffer, oid: Uint8Array): Uint8Array => {
+/**
+ * Wraps the given `payload` in a DER encoding tagged with the given encoded `oid` like so:
+ * `SEQUENCE(oid, BITSTRING(payload))`
+ *
+ * @param paylod The payload to encode as the bit string
+ * @param oid The DER encoded (and SEQUENCE wrapped!) OID to tag the payload with
+ */
+export const wrapDER = (payload: ArrayBuffer, oid: Uint8Array): Uint8Array => {
   // The Bit String header needs to include the unused bit count byte in its length
-  const bitStringHeaderLength = 2 + encodeLenBytes(cose.byteLength + 1);
-  const len = oid.byteLength + bitStringHeaderLength + cose.byteLength;
+  const bitStringHeaderLength = 2 + encodeLenBytes(payload.byteLength + 1);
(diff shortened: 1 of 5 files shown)


packages/identity/src/identity/ed25519.ts
```diff
@@ -27,53 +28,20 @@ export class Ed25519PublicKey implements PublicKey {
...
-    ...[0], // 'no padding'
-  ]);
-
   private static derEncode(publicKey: BinaryBlob): DerEncodedBlob {
-    if (publicKey.byteLength !== Ed25519PublicKey.RAW_KEY_LENGTH) {
-      const bl = publicKey.byteLength;
-      throw new TypeError(
-        `ed25519 public key must be ${Ed25519PublicKey.RAW_KEY_LENGTH} bytes long (is ${bl})`,
-      );
+    // TODO: Validation before encoding is weird, is this necessary?
```
kritzcreek, line 32: Opinions?

packages/identity/src/identity/der.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export const ED25519_OID = Uint8Array.from([
+  ...[0x30, 0x05], // SEQUENCE
+  ...[0x06, 0x03], // OID with 3 bytes
+  ...[0x2b, 0x65, 0x70], // id-Ed25519 OID
+]);
+
+export const wrapDER = (cose: ArrayBuffer, oid: Uint8Array): Uint8Array => {
+  // The Bit String header needs to include the unused bit count byte in its length
+  const bitStringHeader = 2 + encodeLenBytes(cose.byteLength + 1);
```
Dfinity-Bjoern, line 72: That is a length, right? May make sense to reflect that in the name.

packages/identity/src/identity/der.ts
```diff
@@ -0,0 +1,113 @@
...
+const decodeLenBytes = (buf: Uint8Array, offset: number): number => {
+  if (buf[offset] < 0x80) return 1;
+  if (buf[offset] === 0x80) throw new Error('Invalid length 0');
+  if (buf[offset] === 0x81) return 2;
+  if (buf[offset] === 0x82) return 3;
+  if (buf[offset] === 0x83) return 4;
+  throw new Error('Length too long (> 4 bytes)');
+};
+
+export const DER_COSE_OID = Uint8Array.from([
```
Dfinity-Bjoern, line 58: Your code should do the correct thing, so this is just a note re: naming. The OID is technically only the part starting with `0x06` – the `0x03 0x0c` is another `SEQUENCE` wrapping the OID.

The reason for the `SEQUENCE` is that some OIDs come with an additional parameter; the parameter just happens to be empty for both Ed25519 and DER-wrapper-COSE. For other schemes, that is different. For ECDSA, the OID defining the scheme requires an additional OID defining the elliptic curve used.

What I want to say: you are encoding `SEQUENCE(OID)` here, and for ECDSA what you would see is `SEQUENCE(OID, OID)` – so the `SEQUENCE` is not part of the `OID`. Makes sense?

packages/identity/src/identity/ed25519.ts
Dfinity-Bjoern: I think the check could be seen as making sense for the restricted encoding that could only handle fixed-length byte strings. I think it does not make sense anymore. I would then probably also delete the corresponding test.

Dfinity-Bjoern (approved):
In general looks good, one command related to SEQUENCE vs. OID.

packages/identity/src/identity/der.ts
kritzcreek: Right. I figured the extra nuance doesn't mean much outside of this module, as you'd never tease the encoded OID apart for our use-cases so I didn't include it in the name/api. Think I should add your comment as an explanation into the source anyway?

krpeacock (changes requested):
The content of this PR looks good to me. I encourage you to write typedoc comments for the new exports in `der.ts`

kritzcreek:
@krpeacock I added type docs. PTAL

krpeacock (approved):

PR #464, opened 2021-06-30
feat: Enhancements to `LedgerIdentity`.

ielashi:
This PR contains a number of updates to `LedgerIdentity` that were
necessary for the integration with the NNS dapp.

* Calls to the identity automatically open and close the connection to
  the device. Previously, the connection remained open, and there was no
  way to manage it from outside the identity class.

* LedgerIdentity now verifies that the public key of the wallet is the
  public key of the identity. That way, a LedgerIdentity can only
  be used with the wallet that it was instantiated with.

* Additional error-handling has been added.


krpeacock:
closing due to ledger identity no longer being supported in this project

PR #476, opened 2021-07-29
refactor: move all BLOB and Buffer to ArrayBuffer (breaking change)

hansl:
High level changelog:

- This PR removes ALL usage of `Buffer` and (the internal type) `BinaryBlob`. People are now supposed to use the native type `ArrayBuffer`.
- This removes the need to add `fallback` and `ProvidePlugin` for Buffer specifically in the `webpack.config.js` of applications (as seen in the service worker and demos).
  - The only exception is if a third library requires it. For example, `ledgerhq` requires using `Buffer` so using `@dfinity/identity-ledgerhq` will require the Buffer type to be in the `webpack.config.js`.
- Retyped `Certificate.prototype.lookup` parameter type with the more lax version which takes a string. This cleans up nicely now that `lookup` takes `ArrayBuffer`. Removed `lookupEx` function which is now useless (just use `lookup`).
- Moved back `makeNonce` and `Nonce` into `@dfinity/agent`. No reason they should ever be in candid.
- The default `Nonce` value now is just a u128 of the BigInt taken from time + random value. It only needs to be unique, doesn't have to be complicated (and `lebEncode` is probably very slow).
- The way `RequestId` is calculated means that people might end up getting errors for unknown types if they try to calculate a RequestId with a Buffer (say if they mismatch the versions).
- Builds might fail as `BinaryBlob` and its friends have been completely removed (not just retyped)
- Performance should be improved across the board in general, but there's no benchmark in this repo so it's hard to put numbers on it.
- A new `Signature` type has been added that identities need to return now (and cast to). It's a branded `ArrayBuffer`.

Developer notes:
- Adding `@jest-environment node` to some unit tests that require `ArrayBuffer`. These tend to conflict with some tests that require `jsdom` as the jest environment, so we only put it on test files that require it.
- Some utilities (`concat`, `toHex`, `fromHex` for example) have been copy pasted around. They're small and aren't exported, but it's unclear where they should live so I left them duplicated where needed.
- Some quality of life changes made it in, like reordering imports in some files, and adding comments to some undocumented functions. Nothing consistent, just leaving the repo better than I found it.

More work should be done as followup (that I won't have time):
- [ ] Remove others provided things in applications' webpack config (`process`, `assert`, etc). Some of those are in BIP39 for example, which would require replacing it.
- [ ] Resolve the `jsdom` vs `node` environment in jest. Agent JS should remove node AND DOM typings. But using `jsdom` environment makes it weird to use `ArrayBuffer` (as `new ArrayBuffer() instanceof ArrayBuffer` is false), and using `node` environment seems to have the wrong class for `Response` from `fetch` (it doesn't have `arrayBuffer()`). Solution is to implement a jest environment specifically for this project, which is out of scope of this PR.
- [ ] Replace `borc` with something that uses `ArrayBuffer` natively, or replace `simple-cbor` with a better `borc` encoder (subclassing like in this PR).


chenyan-dfinity (approved):
LGTM on the Candid side.

kritzcreek (commented):
Do we have any benchmarks for the candid implementation?

I'm very excited about the API change, but worried this will be a big performance degradation

packages/candid/src/utils/buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+    this._view = this._view.subarray(1);
+    return result;
+  }
+
+  /**
+   * Write a buffer to the end of the pipe.
+   * @param buf The bytes to write.
+   */
+  public write(buf: ArrayBuffer): void {
+    this._view = new Uint8Array([...this._view, ...new Uint8Array(buf)]);
```
kritzcreek, line 74: This creates a full copy of the `_view` array, and it's used in pretty tight loops for individual bytes.

hansl: We could have a better implementation for this, where the `_view` is a subarray of the buffer, and the buffer grows exponentially. This would lead to amortized linear performance instead of polynomial. I can do that early next week if @krpeacock wants me to.

hansl: Well because I hate myself I decided to bite the bullet and do it. Added a new commit.

krpeacock:
> Do we have any benchmarks for the candid implementation?
> 
> I'm very excited about the API change, but worried this will be a big performance degradation

@kritzcreek We don't have benchmarks to compare. The closest example we would have might be the execution time of our e2e tests. Maybe we could run the tests 20 times before and after, and compare the average performance?

hansl:
@krpeacock You could use something like https://www.npmjs.com/package/jest-bench too, quite easily. It's outside the scope of this PR though.

kritzcreek:
> Maybe we could run the tests 20 times before and after, and compare the average performance?

I was primarily worried about the encoding speed of the JS candid implementation. Do any of the E2E tests send some significant data (1-2Mb)? If they do, we might see some change, but if they don't it'll all be lost in the noise.

If dead19e does what it says on the tin my main worry is gone, that change just jumped out at me.

krpeacock:
We can test the changes, but I'm going to go ahead and merge. `0.9.3` will be the stable version, and these changes can go out as `0.10.0-beta.1`

PR #483, opened 2021-08-10
 [WIP] Secp256k1Identity

rocky-fleek:
## Summary
Added `Secp256k1Identity` and `Secp256k1PublicKey` classes and corresponding tests. 
The public key class was taken from the `ledger_hq` code.
The identity class is modeled after the `Ed25519Identity` class and tries to replicate all of its behavior.

### DIFF
--- a/packages/identity/src/identity/secp256k1.ts
+++ b/packages/identity/src/identity/secp256k1.ts
@@ -8,9 +8,9 @@ import {
   blobToHex,
   BinaryBlob,
   derBlobFromBlob,
-  blobFromBuffer,
+  DerEncodedBlob,
 } from '@dfinity/candid';
-import { PublicKey, SignIdentity, DerEncodedBlob } from '@dfinity/agent';
+import { PublicKey, SignIdentity } from '@dfinity/agent';
 import { randomBytes } from 'crypto';
 
 declare type PublicKeyHex = string;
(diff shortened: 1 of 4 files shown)


packages/identity/src/identity/secp256k1.ts
```diff
@@ -0,0 +1,232 @@
...
+export class Secp256k1KeyIdentity extends SignIdentity {
+  public static generate(seed?: Uint8Array): Secp256k1KeyIdentity {
+    if (seed && seed.length !== 32) {
+      throw new Error('Secp256k1 Seed needs to be 32 bytes long.');
+    }
+    // TODO: Add seed parameter --> derive PK from it
+    let privateKey = seed || randomBytes(32); // TODO: REMOVE THIS SINCE IT'S OVERWRITING THE SEED
+    while (!Secp256k1.privateKeyVerify(privateKey)) {
+      privateKey = randomBytes(32);
+    }
```
rocky-fleek, line 119: Need to fix this.

packages/identity/src/identity/secp256k1.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  public getPem(): string {
+    const rawPrivateKey = this._privateKey.toString('hex');
+    const rawPublicKey = this._publicKey.toRaw().toString('hex');
+
+    return `${PEM_BEGIN}\n${Buffer.from(
+      `${PRIV_KEY_INIT}${rawPrivateKey}${KEY_SEPARATOR}${rawPublicKey}`,
+      'hex',
+    ).toString('base64')}\n${PEM_END}`;
+  }
```
rocky-fleek, line 218: @robin-kunzler we added this feature to export the identity in PEM format, to be imported into dfx. 
Let me know if you want this removed to keep the functionality consistent with the ed25519 identity.

krpeacock: I think I would remove the PEM export, since we expect this to run in browsers, and that could potentially lead to some security issues

krpeacock:
Thanks for opening this! I'll take over and see this through

tomiir:
> Thanks for opening this! I'll take over and see this through

Thank you! Let me know if I can give a hand with anything. Also, I can provide the script used to generate the test vectors for the DER encoding in case you need it!

kritzcreek:
I'd suggest trying if https://github.com/dfinity/agent-js/blob/main/packages/identity/src/identity/der.ts can be used to clean up the DER wrapping/unwrapping and make it more robust.

krpeacock:
@kritzcreek I attempted that over in #484. It turns out that the DER encoding pattern can't be simply used for secp without refactoring

krpeacock:
Closing after this functionality was finalized in #484 . Thank you for your contribution!

PR #484, opened 2021-08-10
feat: secp256k1 support in agent js

krpeacock:
[SDK-81], references [483]

### DIFF
--- a/packages/identity/src/buffer.ts
+++ b/packages/identity/src/buffer.ts
@@ -13,3 +13,18 @@ export function fromHexString(hexString: string): ArrayBuffer {
 export function toHexString(bytes: ArrayBuffer): string {
   return new Uint8Array(bytes).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
 }
+
+export function compare(b1: ArrayBuffer, b2: ArrayBuffer): number {
+  if (b1.byteLength !== b2.byteLength) {
+    return b1.byteLength - b2.byteLength;
+  }
+
+  const u1 = new Uint8Array(b1);
+  const u2 = new Uint8Array(b2);
+  for (let i = 0; i < u1.length; i++) {
+    if (u1[i] !== u2[i]) {
+      return u1[i] - u2[i];
+    }
+  }
+  return 0;
+}
--- a/packages/identity/src/identity/der.ts
+++ b/packages/identity/src/identity/der.ts
@@ -74,6 +74,15 @@ export const ED25519_OID = Uint8Array.from([
 ]);
 
 /**
+ * PLACEHOLDER - DO NOT MERGE THIS
+ */
+export const SECP256K1_OID = Uint8Array.from([
+  ...[0x30, 0x05], // SEQUENCE
+  ...[0x06, 0x03], // OID with 3 bytes
+  ...[0x2b, 0x65, 0x70], // id-Ed25519 OID
+]);
+
+/**
  * Wraps the given `payload` in a DER encoding tagged with the given encoded `oid` like so:
  * `SEQUENCE(oid, BITSTRING(payload))`
  *
--- a/packages/identity/src/identity/secp256k1.test.ts
+++ b/packages/identity/src/identity/secp256k1.test.ts
@@ -1,87 +1,113 @@
-import { DerEncodedPublicKey, PublicKey } from '@dfinity/agent';
-import { fromHexString, toHexString } from '../buffer';
+import { DerEncodedPublicKey } from '@dfinity/agent';
+import { blobFromHex, blobFromUint8Array, blobToUint8Array, DerEncodedBlob } from '@dfinity/candid';
+import { randomBytes } from 'crypto';
+import { sha256 } from 'js-sha256';
+import Secp256k1 from 'secp256k1';
 import { Secp256k1KeyIdentity, Secp256k1PublicKey } from './secp256k1';
 
-import { randomBytes } from 'crypto';
-import secp256k1 from 'secp256k1/elliptic';
+function fromHexString(hexString: string): ArrayBuffer {
+  return new Uint8Array((hexString.match(/.{1,2}/g) ?? []).map(byte => parseInt(byte, 16))).buffer;
+}
 
-const generateRandPubKey = () => {
-  let privKey;
-  do {
-    privKey = Uint8Array.from(randomBytes(32));
-  } while (!secp256k1.privateKeyVerify(privKey));
+// DER KEY SECP256K1 PREFIX = 3056301006072a8648ce3d020106052b8104000a03420004
+const testVectors: Array<[string, string]> = [
+  [
+    '0401ec030acd7d1199f73ae3469329c114944e0693c89502f850bcc6bad397a5956767c79b410c29ac6f587eec84878020fdb54ba002a79b02aa153fe47b6ffd33',
+    '3056301006072a8648ce3d020106052b8104000a0342000401ec030acd7d1199f73ae3469329c114944e0693c89502f850bcc6bad397a5956767c79b410c29ac6f587eec84878020fdb54ba002a79b02aa153fe47b6ffd33',
+  ],
+  [
+    '04eb3cbc81bdfa1d13c410151308b4ac65cc328620e368fe5302e2cf2cb7175132999dc2afff871a68afd66d76dd9c7c3911748e8e54ee2fb71df1a97bb4811c8c',
+    '3056301006072a8648ce3d020106052b8104000a03420004eb3cbc81bdfa1d13c410151308b4ac65cc328620e368fe5302e2cf2cb7175132999dc2afff871a68afd66d76dd9c7c3911748e8e54ee2fb71df1a97bb4811c8c',
+  ],
+  [
+    '04fdac09ea93a1b9b744b5f19f091ada7978ceb2f045875bca8ef9b75fa8061704e76de023c6a23d77a118c5c8d0f5efaf0dbbfcc3702d5590604717f639f6f00d',
+    '3056301006072a8648ce3d020106052b8104000a03420004fdac09ea93a1b9b744b5f19f091ada7978ceb2f045875bca8ef9b75fa8061704e76de023c6a23d77a118c5c8d0f5efaf0dbbfcc3702d5590604717f639f6f00d',
+  ],
+];
 
-  privKey; //?
+test('DER encoding of SECP256K1 keys', async () => {
+  testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
+    const publicKey = Secp256k1PublicKey.fromRaw(fromHexString(rawPublicKeyHex));
+    const expectedDerPublicKey = fromHexString(derEncodedPublicKeyHex);
+    expect(publicKey.toDer()).toEqual(expectedDerPublicKey);
+  });
+});
 
-  const pubKey = secp256k1.publicKeyCreate(privKey); //?
+test.only('DER decoding of SECP256K1 keys', async () => {
+  testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
+    const derPublicKey = fromHexString(derEncodedPublicKeyHex) as DerEncodedPublicKey;
+    const expectedPublicKey = fromHexString(rawPublicKeyHex);
+    expect(new Uint8Array(Secp256k1PublicKey.fromDer(derPublicKey).toRaw())).toEqual(
+      new Uint8Array(expectedPublicKey),
+    );
+  });
+});
 
-  toHexString(pubKey).toUpperCase(); //?
+test('DER decoding of invalid keys', async () => {
+  // Too short.
+  expect(() => {
+    Secp256k1PublicKey.fromDer(
+      fromHexString(
+        '3056301006072a8648ce3d020106052b8104000a0342000401ec030acd7d1199f73ae3469329c114944e0693c89502f850bcc6bad397a5956767c79b410c29ac6f587eec84878020fdb54ba002a79b02aa153fe47b6',
+      ) as DerEncodedBlob,
+    );
+  }).toThrow();
+  // Too long.
+  expect(() => {
+    Secp256k1PublicKey.fromDer(
+      fromHexString(
+        '3056301006072a8648ce3d020106052b8104000a0342000401ec030acd7d1199f73ae3469329c114944e0693c89502f850bcc6bad397a5956767c79b410c29ac6f587eec84878020fdb54ba002a79b02aa153fe47b6ffd33' +
+          '1b42211ce',
+      ) as DerEncodedBlob,
+    );
+  }).toThrow();
 
-  pubKey.length; //?
-};
+  // Invalid DER-encoding.
+  expect(() => {
+    Secp256k1PublicKey.fromDer(
+      fromHexString(
+        '0693c89502f850bcc6bad397a5956767c79b410c29ac6f54fdac09ea93a1b9b744b5f19f091ada7978ceb2f045875bca8ef9b75fa8061704e76de023c6a23d77a118c5c8d0f5efaf0dbbfcc3702d5590604717f639f6f00d',
+      ) as DerEncodedBlob,
+    );
+  }).toThrow();
+});
 
-// ED25519 tests for reference
+test('fails with improper seed', () => {
+  expect(() => Secp256k1KeyIdentity.generate(new Uint8Array(new Array(31).fill(0)))).toThrow();
+  expect(() => Secp256k1KeyIdentity.generate(new Uint8Array(new Array(33).fill(0)))).toThrow();
+});
 
-// test('DER encoding of ED25519 keys', async () => {
-//   testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
-//     const publicKey = Secp256k1PublicKey.fromRaw(fromHexString(rawPublicKeyHex));
-//     const expectedDerPublicKey = fromHexString(derEncodedPublicKeyHex);
-//     expect(publicKey.toDer()).toEqual(expectedDerPublicKey);
-//   });
-// });
+test('same seed generates same identity', () => {
+  const seed = new Uint8Array(randomBytes(32));
+  const id = Secp256k1KeyIdentity.generate(seed);
+  const id2 = Secp256k1KeyIdentity.generate(seed);
+  expect(new Uint8Array(id.getPublicKey().toDer())).toEqual(
+    new Uint8Array(id2.getPublicKey().toDer()),
+  );
+});
 
-// test('DER decoding of ED25519 keys', async () => {
-//   testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
-//     const derPublicKey = fromHexString(derEncodedPublicKeyHex) as DerEncodedPublicKey;
-//     const expectedPublicKey = fromHexString(rawPublicKeyHex);
-//     expect(new Uint8Array(Secp256k1PublicKey.fromDer(derPublicKey).toRaw())).toEqual(
-//       new Uint8Array(expectedPublicKey),
-//     );
-//   });
-// });
+test('can encode and decode to/from JSON', () => {
+  const key = Secp256k1KeyIdentity.generate();
+  const json = JSON.stringify(key);
+  const key2 = Secp256k1KeyIdentity.fromJSON(json);
 
-// test('DER decoding of invalid keys', async () => {
-//   // Too short.
-//   expect(() => {
-//     Secp256k1PublicKey.fromDer(
-//       fromHexString(
-//         '302A300506032B6570032100B3997656BA51FF6DA37B61D8D549EC80717266ECF48FB5DA52B65441263484',
-//       ) as DerEncodedPublicKey,
-//     );
-//   }).toThrow();
-//   // Too long.
-//   expect(() => {
-//     Secp256k1PublicKey.fromDer(
-//       fromHexString(
-//         '302A300506032B6570032100B3997656BA51FF6DA37B61D8D549EC8071726' +
-//           '6ECF48FB5DA52B654412634844C00',
-//       ) as DerEncodedPublicKey,
-//     );
-//   }).toThrow();
+  expect(new Uint8Array(key.getPublicKey().toDer())).toEqual(
+    new Uint8Array(key2.getPublicKey().toDer()),
+  );
+});
 
-//   // Invalid DER-encoding.
-//   expect(() => {
-//     Secp256k1PublicKey.fromDer(
-//       fromHexString(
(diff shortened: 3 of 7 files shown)


kritzcreek:
I've made changes to use the existing DER machinery in https://github.com/krpeacock/agent-js/pull/1. I'll take a proper look at the other changes tomorrow.

Dfinity-Bjoern (commented):
Thanks, @krpeacock! I've left a few specific comments. I'll also add @robin-kunzler as a reviewer, since he has a lot more experience with crypto code reviews.

One major application of secp256k1 keys is certainly for keys derived via BIP-32 from seed phrases, such as in Plug. Do we want to add key derivation for these keys as well in the agent? (I think we do have it for Ed25519.)

packages/identity/src/identity/der.ts
```diff
@@ -55,6 +55,16 @@ const decodeLenBytes = (buf: Uint8Array, offset: number): number => {
   throw new Error('Length too long (> 4 bytes)');
 };
 
+const decodeLen = (buf: Uint8Array, offset: number): number => {
+  const lenBytes = decodeLenBytes(buf, offset);
+  if (lenBytes === 1) return buf[offset];
+  else if (lenBytes === 2) return buf[offset + 1];
+  else if (lenBytes === 3) return buf[offset + 1] << (8 + buf[offset + 2]);
```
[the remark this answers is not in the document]
kritzcreek: Good catch! We don't, and the formatter inserted the parenthesis on commit so I never ended up seeing them. I suspect we're not seeing any lengths with more than 1 byte, which is why we're not hitting the bug here.

I should brush up on my JS operator precedences ;)

robin-kunzler:
@krpeacock : Thanks for the PR! FYI I won't be able to do a review today, but happy to do it tomorrow!

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+import { sha256 } from 'js-sha256';
+import Secp256k1 from 'secp256k1';
+import { Secp256k1KeyIdentity, Secp256k1PublicKey } from './secp256k1';
+
+function fromHexString(hexString: string): ArrayBuffer {
+  return new Uint8Array((hexString.match(/.{1,2}/g) ?? []).map(byte => parseInt(byte, 16))).buffer;
+}
+
+// DER KEY SECP256K1 PREFIX = 3056301006072a8648ce3d020106052b8104000a03420004
+const testVectors: Array<[string, string]> = [
```
robin-kunzler, line 12: Would it make sense to add a comment on what the vectors contain and how / where they were obtained?

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+test('DER encoding of SECP256K1 keys', async () => {
+  testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
+    const publicKey = Secp256k1PublicKey.fromRaw(fromHexString(rawPublicKeyHex));
+    const expectedDerPublicKey = fromHexString(derEncodedPublicKeyHex);
+    expect(publicKey.toDer()).toEqual(expectedDerPublicKey);
+  });
+});
+
+test('DER decoding of SECP256K1 keys', async () => {
```
robin-kunzler, line 35: This tests the encoding (toDer) again, exactly as the test above and thus it looks to me like this tests the wrong thing. The test should instead call `fromDer` on the `derEncodedPublicKeyHex` and then compare the raw, right?

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  testVectors.forEach(([rawPublicKeyHex, derEncodedPublicKeyHex]) => {
+    const derEncoded = Secp256k1PublicKey.fromRaw(fromHexString(rawPublicKeyHex)).toDer();
+
+    const testCase = fromHexString(derEncodedPublicKeyHex);
+
+    expect(derEncoded).toEqual(testCase);
+  });
+});
+
+test('DER decoding of invalid keys', async () => {
```
robin-kunzler, line 45: nice that the negative cases are tested!
nit: I'm not a typescript expert, but would it be possible to assert the error message that is thrown?

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+test('can encode and decode to/from JSON', () => {
+  const key = Secp256k1KeyIdentity.generate();
+  const json = JSON.stringify(key);
+
+  const key2 = Secp256k1KeyIdentity.fromJSON(json);
+
+  expect(new Uint8Array(key.getPublicKey().toDer())).toEqual(
+    new Uint8Array(key2.getPublicKey().toDer()),
+  );
```
robin-kunzler, line 96: Since the identity also contains the private key, would it make sense to also make sure they equal? For this, would it be useful to implement equality on the Secp256k1KeyIdentity?

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+});
+
+test('message signature is valid', async () => {
+  const identity = Secp256k1KeyIdentity.generate();
+  const rawPublicKey = identity.getPublicKey().toRaw();
+
+  rawPublicKey.byteLength; //?
```
robin-kunzler, line 103: Nice to have a test that ensures that a valid signature is created!
nit: could this line be removed?

packages/identity/src/identity/secp256k1.test.ts
```diff
@@ -0,0 +1,117 @@
...
+test('same seed generates same identity', () => {
+  const seed = new Uint8Array(randomBytes(32));
+  const id = Secp256k1KeyIdentity.generate(seed);
+  const id2 = Secp256k1KeyIdentity.generate(seed);
+  expect(new Uint8Array(id.getPublicKey().toDer())).toEqual(
+    new Uint8Array(id2.getPublicKey().toDer()),
+  );
+});
+
+test('can encode and decode to/from JSON', () => {
```
robin-kunzler, line 88: maybe add a tests that decoding invalid JSON and JSON with wrong content fails?

packages/identity/src/buffer.ts
```diff
@@ -13,3 +13,23 @@ export function fromHexString(hexString: string): ArrayBuffer {
 export function toHexString(bytes: ArrayBuffer): string {
   return new Uint8Array(bytes).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
 }
+
+/**
+ * Compares two ArrayBuffers and returns 0 if they are identical
+ * @param b1 {ArrayBuffer} Buffer 1
+ * @param b2 {ArrayBuffer} Buffer 2
+ */
+export function compare(b1: ArrayBuffer, b2: ArrayBuffer): number {
```
robin-kunzler, line 22: I don't see this being used anywhere in this PR. Is it needed? If yes, would it make sense to add unit tests for this?

packages/identity/src/identity/secp256k1.ts
```diff
@@ -0,0 +1,149 @@
...
+        return this.fromParsedJson([parsed[0], parsed[1]]);
+      }
+      throw new Error('Deserialization error: JSON must have at least 2 items.');
+    } else if (typeof parsed === 'object' && parsed !== null) {
+      throw new Error('Deprecated JSON format for Ed25519 keys.');
+    }
+    throw new Error(`Deserialization error: Invalid JSON type for string: ${JSON.stringify(json)}`);
+  }
+
+  public static fromKeyPair(publicKey: ArrayBuffer, privateKey: ArrayBuffer): Secp256k1KeyIdentity {
```
robin-kunzler, line 93: nit: maybe add documentation to this method stating that it is the responsibility of the caller to generate the private key in a secure way?

robin-kunzler (commented):
@krpeacock: Thanks a lot for adding the secp256k1 support! Please find a few comments / questions below. Many of them are just minor things. :-)

e2e/node/basic/identity.test.ts
```diff
@@ -15,6 +20,12 @@ function createIdentity(seed: number): SignIdentity {
   return Ed25519KeyIdentity.generate(new Uint8Array(seed1));
 }
 
+function createSecp256k1Identity(seed?: number): SignIdentity {
+  const seed1 = new Array(32).fill(0);
```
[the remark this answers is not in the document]
krpeacock: Making the case for using a seed - wouldn't it allow someone to use an additional tool to turn a seed phrase into an IC-compatible secp256k1Identity? I think that's a use case that groups like Plug would be interested in

e2e/node/basic/identity.test.ts
```diff
@@ -15,6 +20,12 @@ function createIdentity(seed: number): SignIdentity {
...
+  if (seed) seed1[0] = seed;
```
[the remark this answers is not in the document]
krpeacock: You're not - this isn't a particularly realistic test, it just verifies that you can get a different identity by passing a seed, and have it accepted by the replica

packages/identity/src/identity/secp256k1.ts
```diff
@@ -0,0 +1,149 @@
...
+  public static fromJSON(json: string): Secp256k1KeyIdentity {
+    const parsed = JSON.parse(json);
+    if (Array.isArray(parsed)) {
+      if (typeof parsed[0] === 'string' && typeof parsed[1] === 'string') {
+        return this.fromParsedJson([parsed[0], parsed[1]]);
+      }
+      throw new Error('Deserialization error: JSON must have at least 2 items.');
+    } else if (typeof parsed === 'object' && parsed !== null) {
+      throw new Error('Deprecated JSON format for Ed25519 keys.');
+    }
```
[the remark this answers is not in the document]
krpeacock: Yes, this was a duplication from the Ed25519 class. That case was to deprecate for an unsupported type. We can remove both cases for 0.10.0 and just assume it's not part of the API

packages/identity/src/identity/der.ts
```diff
@@ -55,6 +55,16 @@ const decodeLenBytes = (buf: Uint8Array, offset: number): number => {
   throw new Error('Length too long (> 4 bytes)');
 };
 
+const decodeLen = (buf: Uint8Array, offset: number): number => {
```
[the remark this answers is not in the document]
krpeacock: We try to keep these libraries as small as possible, because they end up included in frontend bundles and need to be downloaded on every page load in client applications. If the prefix can be represented accurately with a few hard to understand bytes, we will prefer that over including a full library

packages/identity/src/identity/secp256k1.test.ts
krpeacock: Yes please! Could you suggest a comment?

e2e/node/basic/identity.test.ts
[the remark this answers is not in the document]
robin-kunzler: Also copying my comment below here and deleting it (to have the full context in this thread): 

> Should we omit the optional seed argument, or is it needed somewhere? 
> * It's a bit strange that even though the seed is passed in, the private key could still use some other random bytes (line 64). 
> * It seems safer to me from a security perspective if the seed cannot be provided at all, and the method guarantees that the key generation happens appropriately. WDYT? 

Regarding your comment above: 
* The problem with the current implementation is that it is not guaranteed that the key is actually derived from the seed. The reason is that if `!privateKeyVerify`, the key is still freshly random, and thus for the same seed different keys could be generated. 
* I suggested the simplest solution, which is just not supporting generation from a seed. ;-) But if it is a requirement to support this, see the following: 
* From [the code](https://github.com/cryptocoinjs/secp256k1-node/blob/024fbdad3fb64499e8db8b35246c2f0a36afa8c8/lib/elliptic.js#L76) and [tests](https://github.com/cryptocoinjs/secp256k1-node/blob/024fbdad3fb64499e8db8b35246c2f0a36afa8c8/test/privatekey.js#L4) it seems that the key must be 32 bytes and `n` and `zero` are not allowed. 
* So what we could do is to throw an error in this method in case a seed is given and `!privateKeyVerify`. But this would definitely require a good documentation comment of this method, to make it clear to the caller that some seeds are not accepted. 

WDYT? 

@Dfinity-Bjoern: do you have an opinion on this?

packages/identity/src/identity/secp256k1.test.ts
robin-kunzler: Sure, I assume the keys were generated as I suggested [here](https://github.com/Psychedelic/plug-controller/pull/33#issuecomment-895121742) using OpenSSL.

So the comment could be: 
```
// These test vectors contain the hex encoding of the corresponding raw and DER versions 
// of secp256k1 keys that were generated using OpenSSL as follows: 
// openssl ecparam -name secp256k1 -genkey -noout -out private.ec.key
// openssl ec -in private.ec.key -pubout -outform DER -out ecpubkey.der
// hexdump -ve '1/1 "%.2x"' ecpubkey.der
```

e2e/node/basic/identity.test.ts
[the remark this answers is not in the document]
krpeacock: Based on this feedback, I've removed the secret key and added a test for the `fromSecretKey` method. I'll also add tests to ensure safety in the `fromSecretKey` input content

packages/identity/src/buffer.ts
krpeacock: Yep, seems we didn't need it

packages/identity/src/buffer.ts
```diff
@@ -13,3 +13,23 @@ export function fromHexString(hexString: string): ArrayBuffer {
...
+  if (b1.byteLength !== b2.byteLength) {
+    return b1.byteLength - b2.byteLength;
+  }
+
+  const u1 = new Uint8Array(b1);
+  const u2 = new Uint8Array(b2);
+  for (let i = 0; i < u1.length; i++) {
+    if (u1[i] !== u2[i]) {
+      return u1[i] - u2[i];
```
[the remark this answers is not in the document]
krpeacock: I have removed the function as it didn't end up needing to be used

packages/identity/src/identity/der.ts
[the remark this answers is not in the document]
krpeacock: Tests are added. They're not terribly informative tests, but they will warn us if we break anyting

packages/identity/src/identity/secp256k1.test.ts
krpeacock: yes, it is possible

packages/identity/src/identity/secp256k1.test.ts
krpeacock: It was actually just unnecessary. I'll remove the second one

packages/identity/src/identity/secp256k1.test.ts
krpeacock: I think the existing methods make it possible to evaluate equality, and it's not exactly a feature that anyone has clamored for. I'd say we can consider an equality method on both identities as a future enhancement

packages/identity/src/identity/secp256k1.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  public static fromKeyPair(publicKey: ArrayBuffer, privateKey: ArrayBuffer): Secp256k1KeyIdentity {
+    return new Secp256k1KeyIdentity(Secp256k1PublicKey.fromRaw(publicKey), privateKey);
+  }
+
+  public static fromSecretKey(secretKey: ArrayBuffer): Secp256k1KeyIdentity {
+    const publicKey = Secp256k1.publicKeyCreate(new Uint8Array(secretKey), false);
+    const identity = Secp256k1KeyIdentity.fromKeyPair(publicKey, new Uint8Array(secretKey));
+    return identity;
+  }
```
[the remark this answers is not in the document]
krpeacock: I started writing a test for it, but it turns out that `Secp256k1.publicKeyCreate` already provides us those checks for free. Tweetnacl has a similar error for ED25519

krpeacock:
Feedback incorporated! 

- BIP-32 seed phrase importing 
- Identity comparison

would be new to both ED25519 and Secp2556k1, and I will refrain from adding them for now. 

I have chosen to remove the seed option from `generate` based on feedback, and have updated unit and e2e tests to use the new recommended pattern, by creating the identities using `fromSecretKey`

e2e/node/basic/identity.test.ts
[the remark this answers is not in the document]
robin-kunzler: Thanks for adapting this!

packages/identity/src/identity/der.ts
```diff
@@ -22,7 +22,9 @@ const encodeLenBytes = (len: number): number => {
   }
 };
 
-const encodeLen = (buf: Uint8Array, offset: number, len: number): number => {
+export const encodeLen = (buf: Uint8Array, offset: number, len: number): number => {
+  buf[offset];
```
robin-kunzler, line 26: nit: seems like line 26 can be removed.

packages/identity/src/identity/der.ts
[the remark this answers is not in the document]
robin-kunzler: Great, thanks a lot for adding the DER tests! 👍

e2e/node/basic/identity.test.ts
```diff
@@ -13,17 +13,27 @@ import {
...
-  const seed1 = new Array(32).fill(0);
-  seed1[0] = seed;
-  return Ed25519KeyIdentity.generate(new Uint8Array(seed1));
-}
-
-function createSecp256k1Identity(seed?: number): SignIdentity {
-  const seed1 = new Array(32).fill(0);
-  if (seed) seed1[0] = seed;
-  return Secp256k1KeyIdentity.generate(new Uint8Array(seed1));
+  return Ed25519KeyIdentity.fromSecretKey(fromHexString(seededKeys[seed]));
```
robin-kunzler, line 36: This change here and introducing the `seededKeys` above does not seem to be strictly needed, since for the `Ed25519KeyIdentity`, `generate` still takes a `seed`. Is there a reason to no longer use the seed for the Ed25519 case in the tests? 

From my perspective, having the optional seed in the ed25519 case is reasonable and should probably be kept since `tweetnacl.sign.keyPair.fromSeed(seed)` supports key generation using a seed.  

The seed argument was only removed for the `Secp256k1KeyIdentity`. 

The same comment goes for packages/identity/src/identity/delegation.test.ts, i think those changes would also not be strictly needed, right?

packages/identity/src/identity/ed25519.test.ts
```diff
@@ -64,14 +64,29 @@ test('DER decoding of invalid keys', async () => {
   }).toThrow();
 });
 
-test('fails with improper seed', () => {
-  expect(() => Ed25519KeyIdentity.generate(new Uint8Array(new Array(31).fill(0)))).toThrow();
-  expect(() => Ed25519KeyIdentity.generate(new Uint8Array(new Array(33).fill(0)))).toThrow();
+test('fromSecretKey should generate an identity', () => {
```
robin-kunzler, line 67: Thanks for adding these tests!

Should we keep the old test `fails with improper seed` since the seed is still supported for Ed25519?

packages/identity/src/identity/der.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    buf[1] = 5;
+
+    expect(decodeLen(buf, 0)).toBe(5);
+  });
+  test('encoded length of 3', () => {
+    const buf = Uint8Array.from(new Array(10));
+    buf[0] = 130;
+    // returns value stored at index 2
+    buf[2] = 5;
+    expect(decodeLen(buf, 0)).toBe(5);
```
robin-kunzler, line 146: To test the addition that happens here, maybe switch to the following and remove the comment? 
```suggestion
    buf[1] = 1;
    buf[2] = 1;
    expect(decodeLen(buf, 0)).toBe(257);
```

robin-kunzler (commented):
@krpeacock : Thanks a lot for addressing all the comments / questions. This looks very good. It's great to have the DER tests!!

I have a question regarding the tests for `Ed25519KeyIdentity.generate`, see below. Apart from that, I only have minor things.

e2e/node/basic/identity.test.ts
krpeacock: @robin-kunzler I refactored Ed25519KeyIdentity.generate to no longer take a seed. The same rationale we used for not accepting a seed in Secp256k1KeyIdentity seems like it applies there too, and I prefer to have consistent interfaces / patterns across the the package

packages/identity/src/identity/der.test.ts
krpeacock: That is not a valid case 
image

e2e/node/basic/identity.test.ts
robin-kunzler: > Sorry about the confusion - I accidentally reverted that file during development, but I'm pushing up the updated Ed25519KeyIdentity file now

No problem. I see, thanks for clarifying. 

> have Ed25519KeyIdentity.generate to no longer take a seed. The same rationale we used for not accepting a seed in Secp256k1KeyIdentity seems like it applies there too, and I prefer to have consistent interfaces / patterns across the the package.

* I agree that it is nice having the same API. 
* However, in the ed25519 case tweetnacl actually supports key generation from a seed (as opposed to secp256k1). So it would be OK to do it IMO. 
* If the `generate(seed)` variant is just used for testing, I agree that the changed API without seed is better. Also from a security perspective, it's better to have the key generated randomly.
* But if someone used generate with seed, they would now have to switch to the following. That's probably not so nice since they would need to call tweetnacl directly. 
```
const { publicKey, secretKey } = tweetnacl.sign.keyPair.fromSeed(seed);
Ed25519KeyIdentity.fromKeyPair(publicKey, secretKey);
```

If we know that people are currently using `generate` with seed, I'd vote for keeping `generate` with seed. Otherwise I'm completely fine with the change. Maybe @Dfinity-Bjoern has an opinion or input on this?

packages/identity/src/identity/der.test.ts
robin-kunzler: I think you are just missing the initialization at position 0. This test is successful for me: 
```
  test('encoded length of 3', () => {
    const buf = Uint8Array.from(new Array(10));
    buf[0] = 130;
    buf[1] = 1;
    buf[2] = 1;
    expect(decodeLen(buf, 0)).toBe(257);
  });
```

robin-kunzler (commented):
@krpeacock : Thanks for addressing everything, this looks good to me!

I'll quickly wait with my approval until we get Björn's input on [the seed for Ed25519](https://github.com/dfinity/agent-js/pull/484#discussion_r694674686). 

Note that there are still a few comments that you may want to address, but these are all optional from my perspective. [this](https://github.com/dfinity/agent-js/pull/484#discussion_r694638060), [this](https://github.com/dfinity/agent-js/pull/484#discussion_r694649147) , [this](https://github.com/dfinity/agent-js/pull/484#discussion_r694680606), [this](https://github.com/dfinity/agent-js/pull/484#discussion_r694694835)

PR #490, opened 2021-09-10
fix: auth-client identity option

krpeacock:
I encountered a bug trying to intialize an authClient using a Plug identity, where I was still getting an anonymous identity back. This will properly use the option we pass to `AuthClient.create`


lwshang (approved):
LGTM

PR #493, opened 2021-09-23
feat: Add support for /_/raw/ to enable sub-scope service workers.

jplevyak:
Signed-off-by: John Plevyak <jplevyak@gmail.com>

### DIFF
--- a/apps/sw-cert/src/sw/http_request.ts
+++ b/apps/sw-cert/src/sw/http_request.ts
@@ -212,7 +212,7 @@ export async function handleRequest(request: Request): Promise<Response> {
   /**
    * If this is to /_/raw/ then we pass it through as a raw HTTP request.
    */
-  if (!url.hostname.endsWith(swDomains)) {
+  if (url.pathname.startsWith('/_/raw/')) {
     // The certificate is checked by icx-proxy.
     return await fetch(request);
   }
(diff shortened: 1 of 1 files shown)


nomeata:
Is there some more information what “sub-scope service workers” are? I wonder if such pass though paths endanger the security of the whole thing (see https://forum.dfinity.org/t/a-single-node-provider-might-be-able-to-take-over-internet-identities-it-seems/7464, I would love to read your take on that there)

jplevyak:
@nomeata I closed this because it is unnecessary.  Installing of service workers on subscopes bypasses other service workers (for security reasons).  The problem is that if we inject the service worker for all URL accesses other than /api/... then we can't load a service worker for any subscope which prevents dapps from using say Notifications.   The solution is to not inject the service worker at the front end for /_/raw/... but we don't need to make this change since this service worker is already bypassed during registration:

if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {
    navigator.serviceWorker.register('/_/raw/sw.js').then(function(registration) {
      // Registration was successful
      console.log('ServiceWorker registration successful with scope: ', registration.scope);
    }, function(err) {
      // registration failed :(
      console.log('ServiceWorker registration failed: ', err);
    });
  });
}

nomeata:
Ok, fine for here, but _any_ unverified bypass is a security issue … let’s may continue on the forum.

Issue #500, opened 2021-09-29
agent request_id hashValue throws error


plitzenberger:
When using `@dfinity/agent` within a shared npm module, authenticated calls to the backend fail with the error message:

```
Error: Attempt to hash a value of unsupported type: <canister_id>
```

The problem is that [the condition](https://github.com/dfinity/agent-js/blob/2fe3dd9/packages/agent/src/request_id.ts#L33):

```
else if (value instanceof Principal) {
    return hash(value.toUint8Array());
}
```

returns falsy when the value and the Class being compared reach across npm dependencies. It seems to be the case that we end up with two unique initializations of the Principal class. Javascript object equality is based on location in memory, not a structural comparison. [see here](https://forum.dfinity.org/t/mismatching-dfinity-agent-versions-can-cause-hashing-issues/7359)

aramakme:
Close this one when handled as well -> https://github.com/dfinity/agent-js/issues/496

PR #501, opened 2021-09-29
fix: request_id hashValue throws error (#500)

plitzenberger:


jorgenbuilder:
The ‘_isPrincipal’ param exists on Principal. Wouldn’t that be simplest?

 -    else if (value instanceof Principal) {
 +    else if (value._isPrincipal) {

krpeacock:
I like where this is going, but I think there are some scenarios where this could break (despite our lack of unit test coverage for this section). I'll see if I can get an alternative strategy working  with `Object.hasOwnProperty`

krpeacock:
My solution is tracked in #502

Issue #497, opened 2021-09-28
PrincipalClass encodeValue is breaking compatibility with old Principal


aleacevedo:
# PrincipalClass encodeValue is breaking compatibility with old Principal

## Summary

If you are using old Principal library with new Agent library, when you call a method with a Principal like parameter, it breaks. So you may wrap the principal into Principal of the new library. 

https://github.com/dfinity/agent-js/blob/7102eb715dc026d90ebd3b86b11adeaa4e2fde9f/packages/candid/src/idl.ts#L1142

krpeacock:
This was not actually resolved by #502

anthonymq:
Is there a quick fix for it ? @krpeacock

krpeacock:
I'm not sure how pressing of a priority it is to support calling across different minor versions of these packages, when the update to principal was intended to be a breaking change

krpeacock:
If you need the type guard, I recommend using `Principal.from` on the different version to bring it into your preferred package version. We'll be addressing this in an overall backwards compatibility epic

PR #502, opened 2021-09-29
chore: resolves a bug caused by checking instanceof the Principal class

krpeacock:
SDK-148


ericswanson-dfinity (approved):

aleacevedo:
Issue [#497](https://github.com/dfinity/agent-js/issues/497) was closed by this PR. But I suspect it not gonna fix it. What is more, old Principal gonna break this too because old principal doesn't have .toUint8Array method

PR #526, opened 2022-02-04
feat: notify user interruption

### LINKED ISSUE #525: Notify the user interruption as an error
After calling `authClient.login(...)`, if the user closed the Internet Identity tab accidentally or intentionally, the app don't have an elegant way to catch it. The app may stuck at an "authenticating" stage.

PR #526 
Let the `authClient` poll the `closed` state of `this._idpWindow`.
If it's closed by user, notify the caller through `option.onError`.

**Describe alternatives you've considered**
There're multiple work arounds for the app developers
1. Asking the user to refresh the page manually if it looks stuck
2. Keeping the "login" button available until the login succeeds (no "authenticating" stage)
3. Polling the Internet Identity window state through the `authClient._idpWindow` private property (ignore the TypeScript errors)

1 & 2 harm the UX, 3 is hack. It's better for `AuthClient` to encapsulate the polling.

There're also implementation options
1. Polling or messaging
  I suggest polling. An alternative solution is let II send a message on `beforeunload`, but it's not that reliable. The II window may crash by a browser failure or killed by the user, and there's no chance to signal `beforeunload`.
2. Notifying through `onError` or a new callback
  I prefer to reuse the `onError`. If the II window is closed by user, the authentication is interrupted. There should only be 2 terminal state of the auth process, authenticated (handled by `onSuccess`) or not (handled by `onError`). Adding a new callback adds unnecessary complexity to the caller's logic.

**Additional context**
My forum post about this issue
https://forum.dfinity.org/t/how-to-handle-user-interruption-in-internet-identity-auth/10675

wewei:
Closes #525 
This is part of the feature proposal. I'll add necessary comments and test case if the feature is accepted.


packages/auth-client/src/index.ts
```diff
@@ -284,6 +289,23 @@ export class AuthClient {
 
     // Open a new window with the IDP provider.
     this._idpWindow = window.open(identityProviderUrl.toString(), 'idpWindow') ?? undefined;
+
+    // Check if the _idpWindow is closed by user.
+    this._checkUserInterrupt(() => this._handleFailure(ERROR_USER_INTERRUPT, options?.onError));
+  }
+
+  private _checkUserInterrupt(onInterrupt: () => void) {
```
krpeacock, line 297: Since this is used in only one case, I would probably avoid passing the `onInterrupt` callback here and consolidate the `_handleFailure` logic into `_checkUserInterrupt`. Passing onInterrupt to a private method is probably a premature abstraction.

Otherwise, this is a solid proposal, and will be a nice hook for UI's

krpeacock (changes requested):
see comment above

packages/auth-client/src/index.ts
wewei: Thanks for reviewing. I've updated according to your comment.

One more test case is added.
I also add a line in the login success case
```typescript
it('should call onSuccess if recieved a valid success message', async () => {
 ...
 await client.logout();
});
```
otherwise, the following test cases would throw exception `Error: Invalid public key.`

wewei:
@krpeacock ping~

packages/auth-client/src/index.ts
```diff
@@ -284,6 +289,19 @@ export class AuthClient {
...
     this._idpWindow = window.open(identityProviderUrl.toString(), 'idpWindow') ?? undefined;
+
+    // Check if the _idpWindow is closed by user.
+    const checkInterruption = (): void => {
+      // The _idpWindow is opened and not yet closed by the client
+      if (this._idpWindow) {
+        if (this._idpWindow.closed) {
+          this._handleFailure(ERROR_USER_INTERRUPT, options?.onError);
+        } else {
+          setTimeout(checkInterruption, INTERRUPT_CHECK_INTERVAL);
```
krpeacock, line 300: Do we clear this interval anywhere?

wewei:
Hi, Kyle,It's strange that I received this Email, but I cannot see your comment on GitHub.Anyway, I can reply to you this way.The "interval" you mentioned is not a real interval. It's registered by `setTimeout` instead of `setInterval`. The callback is called once, no repeat. Once the II window is closed, we come to the `then` clause, no more future task is registered.ThanksWei发自我的手机-------- 原始邮件 --------发件人： Kyle Peacock ***@***.***>日期： 2022年2月13日周日 半夜12:42收件人： dfinity/agent-js ***@***.***>抄送： Wei Wei ***@***.***>, Author ***@***.***>主    题： Re: [dfinity/agent-js] feat: notify user interruption (PR #526)
@krpeacock commented on this pull request.



In packages/auth-client/src/index.ts:
> @@ -284,6 +289,19 @@ export class AuthClient {
 
     // Open a new window with the IDP provider.
     this._idpWindow = window.open(identityProviderUrl.toString(), 'idpWindow') ?? undefined;
+
+    // Check if the _idpWindow is closed by user.
+    const checkInterruption = (): void => {
+      // The _idpWindow is opened and not yet closed by the client
+      if (this._idpWindow) {
+        if (this._idpWindow.closed) {
+          this._handleFailure(ERROR_USER_INTERRUPT, options?.onError);
+        } else {
+          setTimeout(checkInterruption, INTERRUPT_CHECK_INTERVAL);

Do we clear this interval anywhere?

krpeacock:
> It's strange that I received this Email, but I cannot see your comment on GitHub.Anyway

I had made that comment on mobile early in the morning, and I deleted it after looking closer at the code and realizing I was mistaken 😅.

krpeacock (approved):

PR #528, opened 2022-02-11
chore: auth-client login on success async

### LINKED ISSUE #471: auth-client: login onSuccess async

According [README](https://github.com/dfinity/agent-js/tree/main/packages/auth-client) the function `login` should be use as follow:

```
authClient.login({
  onSuccess: async () => {
    // authClient now has an identity
  },
});
```

To the contrary, currently implemented, the function is [async](https://github.com/dfinity/agent-js/blob/579f8aa3fae946a9060bed39ff56c87151098c49/packages/auth-client/src/index.ts#L246) and the callback is [void](https://github.com/dfinity/agent-js/blob/579f8aa3fae946a9060bed39ff56c87151098c49/packages/auth-client/src/index.ts#L43). Therefore the proper example would be:

```
await authClient.login({
  onSuccess: () => {
    // authClient now has an identity
  },
});
```

Nevertheless, it would probably be for the best if both would support async, respectively:

```
await authClient.login({
  onSuccess: await () => {
    // authClient now has an identity
  },
});
```

A description which match the implementation and I suggest also to have both function and callback support async.

wackyleo459:
Auth-client supports async pattern for login onSuccess as well as login onError. 
This change now correctly reflects the Readme guideline:
```
authClient.login({
  onSuccess: async () => {
    // authClient now has an identity
  },
});
```

Closes #471


packages/auth-client/src/index.ts
```diff
@@ -284,6 +284,8 @@ export class AuthClient {
 
     // Open a new window with the IDP provider.
     this._idpWindow = window.open(identityProviderUrl.toString(), 'idpWindow') ?? undefined;
+
+    this._eventHandler;
```
krpeacock, line 288: Was this intentional?

krpeacock (changes requested):

packages/auth-client/src/index.ts
wackyleo459: That was what you had changed, so I'm guessing yes?

krpeacock (approved):
Looks good, thanks!

PR #529, opened 2022-02-11
feat: httpagent should allow its identity to be invalidated

krpeacock:
A step toward idle timeouts in auth-client

### DIFF
--- a/packages/agent/src/agent/http/http.test.ts
+++ b/packages/agent/src/agent/http/http.test.ts
@@ -6,6 +6,8 @@ import { Principal } from '@dfinity/principal';
 import { requestIdOf } from '../../request_id';
 
 import { JSDOM } from 'jsdom';
+import { AnonymousIdentity } from '../..';
+import { AgentError } from '../../errors';
 const { window } = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
 window.fetch = global.fetch;
 global.window = window;
@@ -252,4 +254,39 @@ describe('getDefaultFetch', () => {
   it.skip('should throw an error if window, global, and fetch are not available', () => {
     // TODO: Figure out how to test the self and default case errors
   });
+});
+
+describe('invalidate identity', () => {
+  const mockFetch: jest.Mock = jest.fn((resource, init) => {
+    return Promise.resolve(
+      new Response(null, {
+        status: 200,
+      }),
+    );
+  });
+  it('should allow its identity to be invalidated', () => {
+    const identity = new AnonymousIdentity();
+    const agent = new HttpAgent({ identity, fetch: mockFetch, host: 'http://localhost' });
+    const invalidate = () => agent.invalidateIdentity();
+    expect(invalidate).not.toThrowError();
+  });
+  it('should throw an error instead of making a call if its identity is invalidated', async () => {
+    const canisterId: Principal = Principal.fromText('2chl6-4hpzw-vqaaa-aaaaa-c');
+    const identity = new AnonymousIdentity();
+    const agent = new HttpAgent({ identity, fetch: mockFetch, host: 'http://localhost' });
+    agent.invalidateIdentity();
+    await agent
+      .call(canisterId, {
+        methodName: 'test',
+        arg: new ArrayBuffer(16),
+      })
+      .catch((reason: AgentError) => {
+        expect(reason.message).toBe(
+          "This identity has expired due this application's security policy. Please refresh your authentication.",
+        );
+      });
+  });
+});
+describe('replace identity', () => {
+  it.todo('should allow an actor to replace its identity');
 });
--- a/packages/agent/src/agent/http/http.test.ts
+++ b/packages/agent/src/agent/http/http.test.ts
@@ -6,7 +6,7 @@ import { Principal } from '@dfinity/principal';
 import { requestIdOf } from '../../request_id';
 
 import { JSDOM } from 'jsdom';
-import { AnonymousIdentity, Identity, SignIdentity } from '../..';
+import { AnonymousIdentity, SignIdentity } from '../..';
 import { Ed25519KeyIdentity } from '../../../../identity/src/identity/ed25519';
 import { AgentError } from '../../errors';
 const { window } = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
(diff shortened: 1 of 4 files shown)


packages/agent/src/agent/http/index.ts
```diff
@@ -194,7 +206,12 @@ export class HttpAgent implements Agent {
     },
     identity?: Identity | Promise<Identity>,
   ): Promise<SubmitResponse> {
-    const id = (await (identity !== undefined ? await identity : await this._identity)) as Identity;
+    const id = await (identity !== undefined ? await identity : await this._identity);
```
lmuntanerDfinity, line 209: Why is `_identity` a Promise?

packages/agent/src/agent/http/index.ts
```diff
@@ -194,7 +206,12 @@ export class HttpAgent implements Agent {
...
+    if (!id) {
+      throw new Error(
```
lmuntanerDfinity, line 211: why not a `IdentityInvalidError`?

packages/agent/src/agent/http/index.ts
```diff
@@ -392,6 +420,14 @@ export class HttpAgent implements Agent {
     return this.rootKey;
   }
 
+  public invalidateIdentity(): void {
```
lmuntanerDfinity, line 423: Is this used only for testing purposes? or is this part of the API to be used somewhere else?

packages/agent/src/agent/http/index.ts
```diff
@@ -392,6 +420,14 @@ export class HttpAgent implements Agent {
...
+    this._identity = null;
+  }
+
+  public replaceIdentity(identity: Identity): void {
```
lmuntanerDfinity, line 427: Is this used only for testing purposes? or is this part of the API to be used somewhere else?

packages/agent/src/errors.ts
```diff
@@ -4,4 +4,9 @@
  *
  * @todo https://github.com/dfinity/agent-js/issues/420
  */
-export class AgentError extends Error {}
+export class AgentError extends Error {
+  constructor(public readonly message: string) {
+    super(message);
+    Object.setPrototypeOf(this, AgentError.prototype);
```
lmuntanerDfinity, line 10: Why this? Isn't it the default behavior when using `new`?

packages/agent/src/agent/http/http.test.ts
```diff
@@ -253,3 +262,95 @@ describe('getDefaultFetch', () => {
...
+    const expectedError =
+      "This identity has expired due this application's security policy. Please refresh your authentication.";
+
+    // Test Agent.call
+    await agent
+      .call(canisterId, {
+        methodName: 'test',
+        arg: new ArrayBuffer(16),
+      })
+      .catch((reason: AgentError) => {
```
lmuntanerDfinity, line 289: why not a `try / catch` block?

lmuntanerDfinity: you could also test that `mockFetch` has not been called

packages/agent/src/agent/http/http.test.ts
```diff
@@ -253,3 +262,95 @@ describe('getDefaultFetch', () => {
...
+});
+describe('replace identity', () => {
+  const mockFetch: jest.Mock = jest.fn();
+  it('should allow an actor to replace its identity', () => {
+    const identity = new AnonymousIdentity();
+    const agent = new HttpAgent({ identity, fetch: mockFetch, host: 'http://localhost' });
+
+    const identity2 = new AnonymousIdentity();
+    const replace = () => agent.replaceIdentity(identity2);
+    expect(replace).not.toThrowError();
```
lmuntanerDfinity, line 319: could you check that the new identity is the new one?

```
expect(agent._identity).toEqual(identity2);
```

packages/agent/src/agent/http/index.ts
robin-kunzler: IIUC this would be used later to trigger a 'logout' if a session expires (implementing an idle timeout), right?

packages/agent/src/agent/http/index.ts
robin-kunzler: I'm also not sure about this: what would `replaceIdentity` be used for?

robin-kunzler (commented):
@krpeacock : Thanks for the PR and for starting the work on the idle timeouts! Looks good to me but I'll leave the detailed JS review to @lmuntanerDfinity. My only question is about the use case regarding replaceIdentity, see below.

packages/agent/src/agent/http/index.ts
krpeacock: Yes - this is the lower level API to support an inactivity logoout

packages/agent/src/agent/http/index.ts
krpeacock: If someone wants to continue using an agent hat has been invalidated, this will allow them to update the same instance with a fresh identity without refreshing the page or re-instantiating a new HttpAgent.

Both of those options are still viable, but this will provide additional flexibility

packages/agent/src/agent/http/http.test.ts
krpeacock: The try catch block was interacting weirdly with this error in Jest. I can definitely check mockFetch though

packages/agent/src/agent/http/http.test.ts
krpeacock: No, the identity is private out of security concerns. It is also cumbersome to try to inspect it out of the mockFetch payload

lmuntanerDfinity: ok!

packages/agent/src/agent/http/http.test.ts
krpeacock: Yep, refactored with these updates

packages/agent/src/agent/http/index.ts
krpeacock: Decisions that predate me - @hansl liked to pre-emptively make API's compatible with async / await, even if it was not required at time of implementation

packages/agent/src/errors.ts
krpeacock: This was from an attempt to verify the error as an `instanceof` the custom error class, but it wasn't successful

lmuntanerDfinity: You can set a static property and an instance variables so that you can:

```
error.kind === AgentError.kind
```

Not as nice, but it's something... hehe

lmuntanerDfinity (approved):
Minor comment, not necessarily to be implemented, just an option. Approving.

PR #531, opened 2022-02-15
fix: decode optional fields

chenyan-dfinity:


krpeacock (approved):
This is an elegant solution. Looks good to me

packages/candid/src/idl.ts
```diff
@@ -860,8 +860,13 @@ export class RecordClass extends ConstructType<Record<string, any>> {
       x[expectKey] = expectType.decodeValue(b, type);
       idx++;
     }
-    if (idx < this._fields.length) {
-      throw new Error('Cannot find field ' + this._fields[idx][0]);
+    for (const [expectKey, expectType] of this._fields.slice(idx)) {
+      if (expectType instanceof OptClass) {
```
nomeata, line 864: Does this also apply to `reserved`?

chenyan-dfinity: Good catch. Let me add it.

packages/candid/src/idl.ts
```diff
@@ -861,7 +861,11 @@ export class RecordClass extends ConstructType<Record<string, any>> {
       idx++;
     }
     for (const [expectKey, expectType] of this._fields.slice(idx)) {
-      if (expectType instanceof OptClass) {
+      if (
+        expectType instanceof OptClass ||
+        expectType instanceof ReservedClass ||
+        expectType instanceof NullClass
```
nomeata, line 867: I think `null` is actually wrong here. This PR is trying to implement this rule, I assume:

```
<nat> not in <fieldtype>;*
record { <fieldtype>;* } <: record { <fieldtype'>;* }
opt empty <: <datatype'>
------------------------------------------------------------------------------
record { <fieldtype>;* } <: record { <nat> : <datatype'>; <fieldtype'>;* }
```

note that the lower bound is `opt empty`, not `null`, so only `opt t` and `reserved` are included. (I’m not saying it would be morally wrong to treat `null` that way, but it’s not what the spec says.)

chenyan-dfinity: Right. Fixed.

PR #535, opened 2022-02-16
chore: upgrading typescript and typedoc

krpeacock:
This blocks publishing the latest docs for 0.10.3

### DIFF
--- a/packages/auth-client/src/index.ts
+++ b/packages/auth-client/src/index.ts
@@ -338,7 +338,7 @@ export class AuthClient {
               );
             }
           } catch (err) {
-            this._handleFailure(err.message, options?.onError);
+            this._handleFailure((err as Error).message, options?.onError);
           }
           break;
         case 'authorize-client-failure':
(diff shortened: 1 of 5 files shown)


packages/auth-client/src/index.ts
```diff
@@ -338,7 +338,7 @@ export class AuthClient {
               );
             }
           } catch (err) {
-            this._handleFailure(err.message, options?.onError);
+            this._handleFailure((err as Error).message, options?.onError);
```
jason-icp, line 341: would this ever not be an `Error` type? would optional chaining be useful here too?

```typescript
this._handleFailure((err as Error)?.message, options?.onError);
```

jason-icp (approved):
Looks good to me. I left one comment inline, although I don't believe that should stop you from merging. There are a number of warnings though, and it may be a good idea to begin addressing them.

packages/auth-client/src/index.ts
krpeacock: It will always be an Error, but Typescript insists that it has to be treated as `undefined`

PR #536, opened 2022-02-17
chore: updating Agent generic interface

krpeacock:
also adds additional comments for new methods and some additional README content.


packages/agent/README.md
```diff
@@ -33,3 +33,33 @@ import { Actor, HttpAgent } from "@dfinity/agent";
...
+
+Use the `Actor.agentOf` method to get the agent of an actor:
+
+```
+const defaultAgent = Actor.agentOf(defaultActor);
+```
+
+This is useful if you need to replace or invalidate the identity used by an actor's agent.
+
+For example, if you want to replace the identity of an actor's agent with a newly authenticated identity from [Internet Identity](https://identity.ic0.app), you can do so by calling the `Actor.replaceAgent` method:
```
lmuntanerDfinity, line 61: Where does this identity comes from in the first place? Is an identity used when creating an actor for the first time?

krpeacock: Yes, an AnonymousIdentity is initialized during the constructor if no other identity is provided

PR #537, opened 2022-02-20
fix: Update webappsec types dependency

jzxchiang1:
Actually, the `CredentialCreationOptions` type from this package is in the [_public interface_](https://github.com/dfinity/agent-js/blob/main/packages/identity/src/identity/webauthn.ts#L133) of the `@dfinity/identity` package, so it should be part of the dependencies, not the devDependencies (because otherwise if it's part of the latter, like right now, it won't get transitively installed by users of `@dfinity/identity`, causing tsc errors for users).

See these links for details:
- https://github.com/Microsoft/types-publisher/issues/81#issuecomment-234051338
- https://stackoverflow.com/a/46011417


jzxchiang1:
Can someone rerun this? I don't know why it's failing.

krpeacock (approved):

krpeacock:
Yeah, the CI issue was unrelated. Thanks for the contribution!

PR #547, opened 2022-03-25
feat: support idle management in AuthClient 

krpeacock:
# Description

Introduces a new IdleManager to detect when a user has gone idle, and calling callbacks 

Fixes SDK-307

# How Has This Been Tested?

- [x] Unit tests
- [x] in krpeacock/auth-client-demo

# Checklist:


krpeacock:
I made the JS docs a little more verbose, but it does a better job of revealing the full API in intellisense
image

packages/auth-client/src/idleManager.ts
```diff
@@ -0,0 +1,136 @@
...
+      let timeout: number | undefined;
+      return (...args: unknown[]) => {
+        // eslint-disable-next-line @typescript-eslint/no-this-alias
+        const context = this;
+        const later = function () {
+          timeout = undefined;
+          func.apply(context, args);
+        };
+        clearTimeout(timeout);
+        timeout = window.setTimeout(later, wait);
```
jason-icp, line 94: is auth-client ever meant to be used in a nodejs environment?

jason-icp (approved):
Nice work. The tests were clear and the code was easy to understand. I left one comment about whether this would ever need to be used in a nodejs environment - it wouldn't support that as-is clearly, but is environment compatibility something we should add to the README?

PR #549, opened 2022-03-30
feat: warn on bad origin

nmattia:
This adds a debug message in case II is replying with a bad origin.

### DIFF
--- a/packages/auth-client/src/index.ts
+++ b/packages/auth-client/src/index.ts
@@ -307,6 +307,7 @@ export class AuthClient {
   private _getEventHandler(identityProviderUrl: URL, options?: AuthClientLoginOptions) {
     return async (event: MessageEvent) => {
       if (event.origin !== identityProviderUrl.origin) {
+        console.log(`WARNING: expected origin '${identityProviderUrl.origin}', got '${event.origin}' (ignoring)`);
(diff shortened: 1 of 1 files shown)


packages/auth-client/src/index.ts
```diff
@@ -307,6 +307,7 @@ export class AuthClient {
   private _getEventHandler(identityProviderUrl: URL, options?: AuthClientLoginOptions) {
     return async (event: MessageEvent) => {
       if (event.origin !== identityProviderUrl.origin) {
+        console.log(`WARNING: expected origin '${identityProviderUrl.origin}', got '${event.origin}' (ignoring)`);
```
krpeacock, line 310:
```suggestion
        console.warn(`WARNING: expected origin '${identityProviderUrl.origin}', got '${event.origin}' (ignoring)`);
```

krpeacock (changes requested):
Actually, can we just switch this to a console.warn? Otherwise looks good

krpeacock (approved):

packages/auth-client/src/index.ts
nmattia: Didn't know of `console.warn`! Thanks

PR #566, opened 2022-04-22
feat: fast ArrayBuffer encoding

ninegua:
# Description

At the moment trying to encode large binary data is very inefficient because one has to convert them to Array first before passing to the IDL encode function.
In fact trying to convert 100MB ArrayBuffer to Array will result in OOM in Node.js.

This patch adds a special case to type check and allows directly passing ArrayBuffer (e.g. Uint8Array) to the encode function and solves the problem. 

# How Has This Been Tested?

A test case is added.

# Checklist:

### DIFF
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -155,10 +155,22 @@ test('IDL encoding (tuple)', () => {
 test('IDL encoding (arraybuffer)', () => {
   // ArrayBuffer, encode only.
   testEncode(
+    IDL.Vec(IDL.Nat8),
+    new Int8Array([0, 1, 2, 3]),
+    '4449444c016d7b01000400010203',
+    'Array of Nat8s',
+  );
+  testEncode(
     IDL.Vec(IDL.Int8),
     new Int8Array([0, 1, 2, 3]),
     '4449444c016d7701000400010203',
-    'Array of Ints',
+    'Array of Int8s',
+  );
+  testEncode(
+    IDL.Vec(IDL.Int16),
+    new Int16Array([0, 1, 2, 3, 32767, -1]),
+    '4449444c016d760100060000010002000300ff7fffff',
+    'Array of Int16s',
   );
   IDL.encode([IDL.Vec(IDL.Nat8)], [new Uint8Array()]);
   IDL.encode([IDL.Vec(IDL.Nat8)], [new Uint8Array(100).fill(42)]);
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -153,7 +153,6 @@ test('IDL encoding (tuple)', () => {
 });
 
 test('IDL encoding (arraybuffer)', () => {
-  // ArrayBuffer, encode only.
   test_(
     IDL.Vec(IDL.Nat8),
     new Uint8Array([0, 1, 2, 3]),
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -154,19 +154,19 @@ test('IDL encoding (tuple)', () => {
 
 test('IDL encoding (arraybuffer)', () => {
   // ArrayBuffer, encode only.
-  testEncode(
+  test_(
     IDL.Vec(IDL.Nat8),
-    new Int8Array([0, 1, 2, 3]),
+    new Uint8Array([0, 1, 2, 3]),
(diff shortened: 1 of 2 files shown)


packages/candid/src/idl.test.ts
```diff
@@ -152,6 +152,20 @@ test('IDL encoding (tuple)', () => {
   );
 });
 
+test('IDL encoding (arraybuffer)', () => {
+  // ArrayBuffer, encode only.
+  testEncode(
```
ninegua, line 157: I didn't change the decoding behavior because it will be a breaking change.

But I will suggest decoding should be changed too. What do people think?

ninegua: Commit [a1e97d6](https://github.com/dfinity/agent-js/pull/566/commits/a1e97d646d00c26f15aa1f78d63559cfeb393775) adds a fast decoding path too. This changes the return type from Array to ArrayBuffers, and will likely break user code when they upgrade.

packages/candid/src/idl.test.ts
```diff
@@ -152,6 +152,38 @@ test('IDL encoding (tuple)', () => {
   );
 });
 
+test('IDL encoding (arraybuffer)', () => {
+  // ArrayBuffer, encode only.
```
chenyan-dfinity, line 156:
```suggestion
```

packages/candid/src/idl.test.ts
```diff
@@ -152,6 +152,38 @@ test('IDL encoding (tuple)', () => {
...
+  );
+  test_(
+    IDL.Vec(IDL.Int16),
+    new Int16Array([0, 1, 2, 3, 32767, -1]),
+    '4449444c016d760100060000010002000300ff7fffff',
+    'Array of Int16s',
+  );
+  test_(
+    IDL.Vec(IDL.Nat64),
+    new BigUint64Array([BigInt(0), BigInt(1), BigInt(1) << BigInt(60), BigInt(13)]),
```
chenyan-dfinity, line 177: Nice! I didn't know `BigUint64Array` exists.

packages/candid/src/idl.ts
```diff
@@ -742,14 +742,20 @@ export class VecClass<T> extends ConstructType<T[]> {
...
-    return Array.isArray(x) && x.every(v => this._type.covariant(v));
+    // Special case for ArrayBuffer
+    const bits = this._type instanceof FixedNatClass ? this._type.bits : (this._type instanceof FixedIntClass ? this._type._bits : 0);
+    return (ArrayBuffer.isView(x) && bits == (x as any).BYTES_PER_ELEMENT * 8)
+           || (Array.isArray(x) && x.every(v => this._type.covariant(v)));
   }
 
   public encodeValue(x: T[]) {
     const len = lebEncode(x.length);
     if (this._blobOptimization) {
```
chenyan-dfinity, line 753: Seems `_blobOptimization` is not needed anymore?

packages/candid/src/idl.ts
```diff
@@ -774,7 +780,37 @@ export class VecClass<T> extends ConstructType<T[]> {
     }
     const len = Number(lebDecode(b));
     if (this._blobOptimization) {
-      return [...new Uint8Array(b.read(len))] as unknown as T[];
+      return new Uint8Array(b.read(len)) as unknown as T[];
+    }
+
+    if (this._type instanceof FixedNatClass) {
+      if (this._type.bits == 8) {
+        return new Uint8Array(b.read(len)) as unknown as T[];
```
chenyan-dfinity, line 788: Is this a breaking change? If so, we need to document it.

packages/candid/src/idl.ts
ninegua: It is still needed if people pass plain array in.

If we remove this, then old code will probably get worse performance unless they also switch from Array to Uint8Array.

packages/candid/src/idl.ts
ninegua: I added a doc string to VecClass. 

Not sure how to add to CHANGELOG.md since there is none in this repo.

chenyan-dfinity (approved):
LGTM

packages/candid/src/idl.ts
chenyan-dfinity: Maybe add a comment in the code. I think eventually, people want to use typedArray for large vector anyway.

ninegua: done

PR #568, opened 2022-05-03
feat: Support WebAuthn Apple Passkeys

dostro:


krpeacock:
@frederikrothenberger can someone from II review this change? Obviously the code is valid, but I don't know how this will affect other devices downstream

frederikrothenberger:
Afaik @nmattia has started to look into this and is now chasing down the administrative changes necessary so that we can accept third party contributions.

nmattia:
> I don't know how this will affect other devices downstream

Same, I'll have more time to look into this next week!

krpeacock:
@nmattia did we ever get a verdict on this? @dostro is this still needed?

nmattia:
@krpeacock I'll admit I don't exactly remember. With iOS 16 I've been offered to authenticate with passkeys and it seemed to work out of the box, but I'm still open to more testing.

PR #570, opened 2022-05-05
feat: idlemanager reload by default

krpeacock:
# Description

To further enhance the IdleManager's default configuration for security, IdleManager now provides a default callback that will refresh the page, fully exiting the session. 

The developer is still encouraged to register their own callback to enhance the user experience. Passing a custom callback will override the default callback

# Checklist:

### DIFF
--- a/packages/auth-client/src/idleManager.test.ts
+++ b/packages/auth-client/src/idleManager.test.ts
@@ -2,6 +2,13 @@ import IdleManager from './idleManager';
 
 jest.useFakeTimers();
 
+const { location } = window;
+
+beforeEach(() => {
+  delete window.location;
+  window.location = { reload: jest.fn() };
+});
+
 describe('IdleManager tests', () => {
   it('should call its callback after time spent inactive', () => {
     const cb = jest.fn();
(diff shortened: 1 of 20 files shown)


packages/auth-client/src/idleManager.test.ts
```diff
@@ -12,6 +19,32 @@ describe('IdleManager tests', () => {
...
+    jest.advanceTimersByTime(10 * 60 * 1000);
+    expect(window.location.reload).toHaveBeenCalled();
+  });
+  it('should replace the default callback if a callback is passed during creation', () => {
+    IdleManager.create({ onIdle: jest.fn() });
+
+    expect(window.location.reload).not.toHaveBeenCalled();
+    // simulate user being inactive for 10 minutes
+    jest.advanceTimersByTime(10 * 60 * 1000);
+    expect(window.location.reload).not.toHaveBeenCalled();
```
dfx-json, line 36: another assertion here could be that the handler passed to `onIdle` actually gets called

dfx-json (approved):

packages/auth-client/src/idleManager.ts
```diff
@@ -114,21 +124,25 @@ class IdleManager {
    * @param {IdleCB} callback function to be called when user goes idle
    */
   public registerCallback(callback: IdleCB): void {
+    if (this._usingDefaultCb) {
+      this.callbacks = [];
```
robin-kunzler, line 128: Do I understand correctly that this is currently not working as intended, since `AuthClient.constructor` will always call `registerCallback(logout)`, actually removing the default callback (the page reload) right after creation? Thus IIUC the logout would never be followed by the reload at the moment. 

Side note: Since we need both the logout and the refresh callbacks, it is important that the logout callback is executed before the refresh.

packages/auth-client/src/idleManager.ts
```diff
@@ -114,21 +124,25 @@ class IdleManager {
...
+    }
```
robin-kunzler, line 129: Would it be possible to add a test that the local storage is cleared AND the refresh happens when the timeout occurs?

robin-kunzler (commented):
Thanks a lot for implementing this right away @krpeacock ! I like the solution of registering also a reload on idle. 

But IIUC there is a bug, see my comments below.

packages/auth-client/src/idleManager.ts
```diff
@@ -26,12 +26,14 @@ const events = ['mousedown', 'mousemove', 'keydown', 'touchstart', 'wheel'];
 
 /**
  * Detects if the user has been idle for a duration of `idleTimeout` ms, and calls `onIdle` and registered callbacks.
- * @see {@link IdleManager}
+ * By default, the IdleManager will log a user out after 30 minutes of inactivity, and reload the page.
```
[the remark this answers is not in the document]
krpeacock: oh, I had this listed wrong in another place too - fortunately the behavior was correct at 10 minutes

packages/auth-client/src/idleManager.ts
krpeacock: Oh, you're right - I'll write a new test for that flow. I'll probably need to update the logic slightly as well

packages/auth-client/src/idleManager.ts
krpeacock: This test will live in the index.test.ts file, since the logout functionality isn't directly a concern of the idleManager itself

robin-kunzler (approved):
@krpeacock : Looks great, thanks! I like the API and the test that verifies that both the storage is cleared and the reload happens. 👍  Approving the PR.

PR #572, opened 2022-05-19
feat: CanisterStatus utility

krpeacock:
# Description
The CanisterStatus utility is used to request structured data directly from the IC public API. This data can be accessed using agent.readState, but CanisterStatus provides a helpful abstraction with some known paths.

You can request a canisters Controllers, ModuleHash, Candid interface, Subnet, or Time, or provide a custom path {@link CanisterStatus.CustomPath} and pass arbitrary buffers for valid paths identified in https://smartcontracts.org/docs/current/references/ic-interface-spec/.

# How Has This Been Tested?

Unit testing & local e2e testing

# Checklist:


packages/agent/src/canisterStatus/index.test.ts
```diff
@@ -0,0 +1,167 @@
...
+};
+
+describe('Canister Status utility', () => {
+  it('should query the time', async () => {
+    const status = await getStatus(['Time']);
+    expect(status.get('Time')).toStrictEqual(new Date('2022-05-18T23:29:38.621Z'));
+  });
+  it('should query canister controllers', async () => {
+    const status = await getStatus(['Controllers']);
+    expect(status.get('Controllers')).toBeTruthy();
```
ericswanson-dfinity, line 78: Is there enough real data in the certificate to verify more than truthiness here?

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,245 @@
+/** @module CanisterStatus */
```
chenyan-dfinity, line 1: Not sure if `canisterStatus` is a good name. `time` and `subnet` are not tied to the canister. Maybe call it `StateTree` to match with the interface spec?

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,245 @@
...
+  key: string;
+  path: string | ArrayBuffer;
+  decodeStrategy: 'cbor' | 'hex' | 'leb128' | 'raw';
+}
+
+/**
+ * Pre-configured fields for canister status paths
+ */
+export type Path =
+  | 'Time'
```
chenyan-dfinity, line 38: lower case to match with spec?

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,245 @@
...
+
+const decodeLeb128 = (buf: ArrayBuffer): bigint => {
+  return lebDecode(new PipeArrayBuffer(buf));
+};
+
+const decodeCbor = (buf: ArrayBuffer): ArrayBuffer[] => {
+  return Cbor.decode(buf);
+};
+
+// Time is a CBOR-encoded Nat
```
chenyan-dfinity, line 232:
```suggestion
// Time is a LEB128-encoded Nat
```

packages/agent/src/certificate.ts
```diff
@@ -119,7 +119,9 @@ export class Certificate {
     const sig = this.cert.signature;
     const key = extractDER(derKey);
     const msg = concat(domain_sep('ic-state-root'), rootHash);
+    blsVerify; //?
```
chenyan-dfinity, line 122: What's this?

packages/agent/src/canisterStatus/index.test.ts
krpeacock: Yes, I can verify the actual controllers

packages/agent/src/canisterStatus/index.ts
krpeacock: My reasoning is that `StateTree` sounds overly generic - state tree of what? The entire IC? 

The feature was motivated by me wanting to discover the ModuleHash, Metadata, Controllers, and Subnet of a canister. I'd like the naming to reflect that motivation, which should make it more discoverable.

the fact that this involves a state tree in a certificate is more or less an implementation detail, from my POV as a developer

packages/agent/src/canisterStatus/index.ts
krpeacock: can do

packages/agent/src/certificate.ts
krpeacock: ah, that's a debugging comment
```suggestion
```

packages/agent/src/certificate.ts
```diff
@@ -120,6 +120,7 @@ export class Certificate {
     const key = extractDER(derKey);
     const msg = concat(domain_sep('ic-state-root'), rootHash);
     const res = await blsVerify(new Uint8Array(key), new Uint8Array(sig), new Uint8Array(msg));
+    res;
```
ericswanson-dfinity, line 123: is this intentionally added?

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Pre-configured fields for canister status paths
+ */
+export type Path =
+  | 'time'
+  | 'controllers'
+  | 'subnet'
+  | 'moduleHash'
```
chenyan-dfinity, line 41:
```suggestion
  | 'module_hash'
```

packages/agent/src/certificate.ts
chenyan-dfinity, line 123:
```suggestion
```

docs/generated/changelog.html
```diff
@@ -16,6 +16,29 @@ <h2>Version 0.11.2</h2>
...
+          Comes with nicely configured options for
+          <pre>Time</pre>
+          ,
+          <pre>Controllers</pre>
+          ,
+          <pre>Subnet</pre>
+          ,
+          <pre>ModuleHash</pre>
+          , and
+          <pre>Candid</pre>
```
chenyan-dfinity, line 36:
```suggestion
          <pre>time</pre>
          ,
          <pre>controllers</pre>
          ,
          <pre>subnet</pre>
          ,
          <pre>module_hash</pre>
          , and
          <pre>candid</pre>
```

packages/agent/src/certificate.ts
krpeacock: nope

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,242 @@
...
+        const cert = new Certificate(response, effectiveAgent);
+        const verified = await cert.verify();
+        if (!verified) {
+          throw new Error(
+            'There was a problem certifying the response data. Please verify your connection to the mainnet, or be sure to call fetchRootKey on your agent if you are developing locally',
+          );
+        }
+
+        const data = cert.lookup(encodePath(uniquePaths[index], canisterId));
+        data;
```
ericswanson-dfinity, line 108: is this for debugging?

krpeacock: Yes, that was for debugging. I have tooling that makes that be the equivalent of a console.log, but it would be a no-op in actual code. Same goes for the special comment `//?` It should still be removed though

packages/agent/src/canisterStatus/index.ts
```diff
@@ -0,0 +1,241 @@
...
+}): Promise<StatusMap> => {
+  const { canisterId, agentOptions, agent, paths } = options;
+
+  const uniquePaths = [...new Set(paths)];
+
+  // Set up agent
+  const effectiveAgent = agent ?? new HttpAgent(agentOptions);
+  if (options.agentOptions?.host && !options.agentOptions.host.startsWith('https://ic0.app')) {
+    await effectiveAgent.fetchRootKey();
+  }
```
ericswanson-dfinity, line 85: 1. If an agent is passed in, I don't think we should ever call fetchRootKey on it.  Instead require the caller to have already set up their agent.
3. Based on this adjustment in the HttpAgent constructor, this comparison doesn't look safe, for example if the host is `<canister-id>.ic0.app`

https://github.com/dfinity/agent-js/blob/main/packages/agent/src/agent/http/index.ts#L186

```
    // Rewrite to avoid redirects
    if (this._host.hostname.endsWith(IC0_SUB_DOMAIN)) {
      this._host.hostname = IC0_DOMAIN;
    }
```

Can we make the caller explicitly indicate that they want to fetch the root key?  If not, would you consider removing the convenience of automatically creating an HttpAgent (so: retain `agent` parameter, remove `agentOptions` parameter)

ericswanson-dfinity (changes requested):

packages/agent/src/canisterStatus/index.ts
krpeacock: agreed. It's a little more cumbersome, but passing an agent in does resolve the potential inconsistency. We also don't manage replica-based behavior anywhere else in the agent, so we shouldn't do it here. I'll remove the convenience constructor

ericswanson-dfinity (approved):

chenyan-dfinity (approved):

PR #577, opened 2022-06-02
fix: add setBigUint64 polyfill

mstrasinskis:
# Description

Adds the `setBigUint64` polyfill because of the multiple [error reports](https://dfinity.slack.com/archives/C01S03NBM7S/p1654161676609929). (based on [nns-dapp solution](https://github.com/dfinity/nns-dapp/blob/55a89a9b5d5d4241277822197efc62f3d3e8d6ce/frontend/ts/src/canisters/converter.ts#L25))

### DIFF
--- a/packages/agent/src/agent/http/http.test.ts
+++ b/packages/agent/src/agent/http/http.test.ts
@@ -362,10 +362,39 @@ describe('replace identity', () => {
     expect(mockFetch).toBeCalledTimes(1);
   });
 });
-describe('makeNonce should create unique values', () => {
-  const nonces = new Set();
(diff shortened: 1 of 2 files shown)


dfx-json (commented):
Hi @mstrasinskis. The proposed changes look good. The only thing I would like to know is how we can verify lines 118 and 119 produce a result equivalent to line 115? Could you perhaps add one more test? Thank you!

bitdivine (approved):

nmattia:
merging on behalf of @bitdivine , apparently GitHub is having a slow day!

PR #580, opened 2022-06-03
fix(agent): Check subnet canister ranges

oggy-dfin:
# Description

We were missing checks that ensured that responses were signed by the correct subnet.

In detail, when the user reads the state of a canister `C`, the IC returns a state tree in a response `Resp`. To ensure the integrity of the `Resp`, the state tree must be signed. It can be signed directly by the root key of the IC, or by a different key (in particular, a subnet key) that is certified by the IC root key through a (possibly nested) delegation. Such delegations are scoped, and valid only for certain canister ranges. For `Resp` to be valid, `C` has to be included in the ranges of all the delegations, but this check was missing. See:

https://internetcomputer.org/docs/current/references/ic-interface-spec/#certification
https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state
https://internetcomputer.org/docs/current/references/ic-interface-spec/#canister-signatures

Fixes SDK-315

# How Has This Been Tested?

Unit tests

# Checklist:


packages/agent/src/canisterStatus/index.ts
```diff
@@ -87,8 +87,8 @@ export const request = async (options: {
         const response = await agent.readState(canisterId, {
           paths: [encodedPaths[index]],
         });
-        const cert = new Certificate(response, agent);
-        const verified = await cert.verify();
+        const cert = new Certificate(response.certificate, agent.fetchRootKey());
```
dfx-json, line 90: `agent.fetchRootKey()` is asynchronous, so this would need an `await` - otherwise, the value that gets passed to the Certificate constructor is a Promise.

```typescript
const rootKey = await agent.fetchRootKey();
const cert = new Certificate(response.certificate, rootKey);
```

packages/agent/src/canisterStatus/index.ts
```diff
@@ -87,8 +87,8 @@ export const request = async (options: {
...
+        const verified = await cert.verify(canisterId);
```
dfx-json, line 91: quick check here - this changes the interface of `verify`. We should bump agent-js to the next minor version instead of the next patch version.

also, is there anywhere else in the codebase that calls `verify()` without arguments?

packages/agent/src/certificate.ts
```diff
@@ -130,28 +129,41 @@ export class Certificate {
...
+      ]);
+
+      const canister_in_range = ranges.some(r => r[0].ltEq(canisterId) && r[1].gtEq(canisterId));
+      if (!canister_in_range) {
+        throw new Error(
+          `Canister ${canisterId} not in range of delegations for subnet 0x${toHex(d.subnet_id)}`,
+        );
+      }
+    }
+    const public_key_lookup = cert.lookup(['subnet', d.subnet_id, 'public_key']);
```
dfx-json, line 162: same nit here re: camelCase

packages/principal/src/index.test.ts
```diff
@@ -24,4 +24,21 @@ describe('Principal', () => {
   it('errors out on parsing invalid characters', () => {
     expect(() => Principal.fromText('Hello world!')).toThrow();
   });
+
+  it('compares principals correctly', () => {
+    const anonymous = Principal.anonymous();
+    const principal1 = Principal.fromText('ryjl3-tyaaa-aaaaa-aaaba-cai');
+    const principal2 = Principal.fromText('ivg37-qiaaa-aaaab-aaaga-cai');
+    for (let p of [anonymous, principal1, principal2]) {
```
dfx-json, line 32: lint error here: the tests are failing because this is a `let` instead of `const`.

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
   public toString(): string {
     return this.toText();
   }
+
+  public compareTo(other: Principal): 'lt' | 'eq' | 'gt' {
+    for (let i = 0; i < Math.min(this._arr.length, other._arr.length); i++) {
+      if (this._arr[i] < other._arr[i]) return 'lt';
+      else if (this._arr[i] > other._arr[i]) return 'gt';
```
dfx-json, line 109: nit: code style - if statements need braces.

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
...
+    }
+    // Here, at least one principal is a prefix of the other principal (they could be the same)
+    if (this._arr.length < other._arr.length) return 'lt';
```
dfx-json, line 112: nit: code style - if statements need braces.

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
...
+    if (this._arr.length > other._arr.length) return 'gt';
```
dfx-json, line 113: nit: code style - if statements need braces.

dfx-json (changes requested):
Nice work @oggy-dfin! I left a few comments and nits. One main comment was around using `await` when calling `fetchRootKey`.

packages/agent/src/certificate.ts
```diff
@@ -130,28 +129,41 @@ export class Certificate {
     }
   }
 
-  private async _checkDelegation(d?: Delegation): Promise<ArrayBuffer> {
+  private async _checkDelegationAndGetKey(
+    canisterId: Principal,
+    d?: Delegation,
+  ): Promise<ArrayBuffer> {
     if (!d) {
-      if (!this._rootKey) {
```
chenyan-dfinity, line 135: Why removing this logic? `fetchRootKey()` should only be called in local replica. The root key for the mainnet is hardcoded and doesn't require explicit fetching.

packages/agent/src/canisterStatus/index.ts
chenyan-dfinity: I have the same concern. Not all certificates contain a path with canister_id. How do we verify the certificate if we only want to read non canister related paths, e.g. `/time`, `/subnet/` and `/request_status`?

chenyan-dfinity (commented):
Note that the same check is missing in `agent-rs`.

packages/agent/src/canisterStatus/index.ts
oggy-dfin: The `Certiificate` constructor takes a `Promise`, it's later awaited on in `Certificate.verify`, where it's used - I didn't want to block before the point where it was needed (not that it makes much of a difference in practice, as the user pretty much has to call `verify` to do anything with the certificate).

packages/agent/src/canisterStatus/index.ts
oggy-dfin: That's a good point - I'll bump the version up. We could also consider rejigging the type of `verify` (or getting rid of it) - see the answer to your next comment.

For the other point (other places calling `verify`), I'd hope that the Typescript type system is sound enough to catch this? However, I did miss one place, revealed by the failed PR tests (I didn't check these, my bad), in `e2e`, because I never called the TS compiler on that package. Let's discuss with Kyle to add a command that builds the whole monorepo in the right order and catches all such errors? I tried to do this as we discussed a couple of weeks ago, but didn't get anywhere in an hour or so that I spent on it.

packages/agent/src/certificate.ts
```diff
@@ -130,28 +129,41 @@ export class Certificate {
...
-    const cert: Certificate = new Certificate(d as any, this._agent);
-    if (!(await cert.verify())) {
+    const cert: Certificate = new Certificate(d.certificate, this._rootKey);
+    if (!(await cert.verify(canisterId))) {
       throw new Error('fail to verify delegation certificate');
     }
 
-    const lookup = cert.lookup(['subnet', d.subnet_id, 'public_key']);
-    if (!lookup) {
+    if (canisterId.compareTo(Principal.managementCanister()) != 'eq') {
```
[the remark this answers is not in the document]
oggy-dfin: If [this post](https://dmitryrogozhny.com/blog/typescript-and-implicit-coercion) is correct, there'd be no coercion in TS (it'd be a compile-time error instead)

packages/agent/src/certificate.ts
```diff
@@ -130,28 +129,41 @@ export class Certificate {
...
+      const range_lookup = cert.lookup(['subnet', d.subnet_id, 'canister_ranges']);
```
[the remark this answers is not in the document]
oggy-dfin: Thanks for catching this - I changed all the `snake_case` I introduced to `camelCase`. However, `subnet_id` was there before, so I didn't touch this - I'm not sure if there's some magic going on because it's parsed from a `subnet_id` field in CBOR.

packages/agent/src/certificate.ts
```diff
@@ -102,20 +102,19 @@ function isBufferEqual(a: ArrayBuffer, b: ArrayBuffer): boolean {
 export class Certificate {
   private readonly cert: Cert;
   private verified = false;
-  private _rootKey: ArrayBuffer | null = null;
 
-  constructor(response: ReadStateResponse, private _agent: Agent = getDefaultAgent()) {
-    this.cert = cbor.decode(new Uint8Array(response.certificate));
+  constructor(certificate: ArrayBuffer, private _rootKey: Promise<ArrayBuffer>) {
```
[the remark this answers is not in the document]
oggy-dfin: If we change the interface to stick the `canisterId` into the certificate that makes total sense - if we for whatever reason want to call `verify` separately (I don't know why we would), a `Promise` would allow to delay blocking until we need it.

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
   public toString(): string {
     return this.toText();
   }
+
+  public compareTo(other: Principal): 'lt' | 'eq' | 'gt' {
+    for (let i = 0; i < Math.min(this._arr.length, other._arr.length); i++) {
+      if (this._arr[i] < other._arr[i]) return 'lt';
```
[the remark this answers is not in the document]
oggy-dfin: If we want to enforce this, can we put it in the linter? I don't think it's good use of time for either of us to be discussing these

packages/agent/src/certificate.ts
oggy-dfin: Thanks for catching this - my fix had made things worse! However, I do think that this is the wrong place for this logic - there's no real need for `Certificate` to know anything about the agent IMO (I ended up staring at this piece of code for a while when first looking it, and obviously got it wrong), it just needs a root key. 

I changed the code now so that the callers use the root key if available. We can probably extract a convenience method on `Agent` to do this.

packages/agent/src/canisterStatus/index.ts
oggy-dfin: For @chenyan-dfinity 's point: any call to `read_state` (including those to get time, subnet and request status) has to have an effective canister ID, and the scope of the certificate must be checked against this ID:
https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state
Is there another way to obtain such a certificate?

oggy-dfin:
> Note that the same check is missing in `agent-rs`.

The check was added to `agent-rs` in https://github.com/dfinity/agent-rs/pull/312

packages/agent/src/certificate.ts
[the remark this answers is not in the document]
chenyan-dfinity: Let's not rely on TS for correctness. If one side the operand is `any` type, the check will pass, and we still get coercion when translated to JS.

packages/agent/src/canisterStatus/index.ts
oggy-dfin: > So we need to at least disable the check for the provisional API.

I think the interface spec is actually wrong on this, as it says that any canister is an effective canister ID for `provisional_create_canister_with_cycles`, but then also requires that the delegation is checked for the call to `read_state`. I think we should disable the check for the case when `aaaaa-aa` is the effective canister ID, as per the spec it can only be the effective canister ID for the `provisional_create_canister_with_cycles` call. I'll discuss this with Jens once he's back from vacation.

> Also, can this effective_canister_id be a member of the Certificate class, so that we don't have to pass the canister id when calling verify? This also makes it a non-breaking change. We can disable the check if the effective_canister_id is missing.

Since we're fixing a security flaw, I think we have to make a breaking change (callers that don't change the call would stay vulnerable). After discussing with Jason, since it's a breaking change already, I've changed the API to perform the verification immediately in the constructor - we couldn't find a use case where you'd want to delay the verification (do you maybe see one?). Also, I unified the error reporting to always use exceptions, instead of a mixture of exceptions and Booleans as was previously the case for `verify`

packages/agent/src/certificate.ts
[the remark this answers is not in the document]
oggy-dfin: OK, changed to `!==`

packages/agent/src/polling/index.ts
```diff
@@ -29,8 +29,9 @@ export async function pollForResponse(
 ): Promise<ArrayBuffer> {
   const path = [new TextEncoder().encode('request_status'), requestId];
   const state = await agent.readState(canisterId, { paths: [path] });
-  const cert = new Certificate(state, agent);
-  const verified = await cert.verify();
+  const rootKey = agent.rootKey == null ? agent.fetchRootKey() : Promise.resolve(agent.rootKey);
+  const cert = new Certificate(state.certificate, rootKey);
+  const verified = await cert.verify(canisterId);
```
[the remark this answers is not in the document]
oggy-dfin: We don't have sufficient information at this point in the code to compute this (there's no request); per the comment, the function already expects an effective canister ID. I don't see any special computation of the effective ID, but I may have missed something.

So I think you're right and we should add it, but I don't think this PR is breaking anything beyond what's broken already. In particular, to read the status of calls to the management canister other than `provisional_create_canister_with_cycles`, you need to compute the effective canister ID regardless of whether you're properly checking the certificate, and then `read_state` on that ID, otherwise (assuming that the replica actually follows the interface spec) you don't get a response.

packages/agent/src/canisterStatus/index.ts
chenyan-dfinity: > I think the interface spec is actually wrong on this
I think the spec is fine. We just need to disable the check when the effective canister id is `aaaaa-aa`. agent-rs does the same thing: https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/mod.rs#L1214

> Since we're fixing a security flaw, I think we have to make a breaking change (callers that don't change the call would stay vulnerable)

Not sure it has to be a breaking change. Callers that didn't upgrade agent-js to the fixed version stays vulnerable. It doesn't have to be a breaking change to fix a vulnerability.

> I've changed the API to perform the verification immediately in the constructor - we couldn't find a use case where you'd want to delay the verification (do you maybe see one?).

There might be a use case that the caller doesn't want to do a verification. If we verify in the constructor, it can be harder to skip verification? The problem is that bls verify is written in Wasm, some platforms don't support wasm execution, e.g. mobile platform, chrome plugins. In those environments, they either have to use a slower JS implementation or skip the verification completely. I know a grant awardee is writing a Chrome plugin to inspect all IC traffic, and cannot run verify in the plugin setting. Since it's a debugging tool, it's not too bad to disable the verification.

krpeacock:
One note as I'm starting to review this - while I agree that this change should go out under `0.12.0`, our release process for this project is to defer bumping package.json versions until we initiate the release, handling that in a dedicate PR. 

I'll go ahead and clean that up, which should also remove some overhead in the changed files view

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
   public toString(): string {
     return this.toText();
   }
+
+  public compareTo(other: Principal): 'lt' | 'eq' | 'gt' {
```
krpeacock, line 106:
```suggestion
  /**
   * Utility method comparing a Principal to compare against. Used for determining canister ranges in certificate verification
   * @param {Principal} other - a {@link Principal} to compare
   * @returns {'lt' | 'eq' | 'gt'} `'lt' | 'eq' | 'gt'` a string, representing less than, equal to, or greater than
   */
  public compareTo(other: Principal): 'lt' | 'eq' | 'gt' {
```

packages/principal/src/index.ts
```diff
@@ -96,4 +102,25 @@ export class Principal {
...
+      if (this._arr[i] < other._arr[i]) return 'lt';
+      else if (this._arr[i] > other._arr[i]) return 'gt';
+    }
+    // Here, at least one principal is a prefix of the other principal (they could be the same)
+    if (this._arr.length < other._arr.length) return 'lt';
+    if (this._arr.length > other._arr.length) return 'gt';
+    return 'eq';
+  }
+
+  public ltEq(other: Principal): boolean {
```
krpeacock, line 117:
```suggestion
  /**
   * Utility method checking whether a provided Principal is less than or equal to the current one using the {@link Principal.compareTo} method
   * @param other a {@link Principal} to compare
   * @returns {boolean} boolean
   */
  public ltEq(other: Principal): boolean {
```

krpeacock (commented):
I've contributed some doc comments for Principal since I try to include them for all new methods and to add more coverage over time

packages/agent/src/certificate.ts
```diff
@@ -101,57 +100,89 @@ function isBufferEqual(a: ArrayBuffer, b: ArrayBuffer): boolean {
...
-  private verified = false;
-  private _rootKey: ArrayBuffer | null = null;
 
-  constructor(response: ReadStateResponse, private _agent: Agent = getDefaultAgent()) {
-    this.cert = cbor.decode(new Uint8Array(response.certificate));
+  /**
+   * Create a new instance of a certificate, automatically verifying it.
+   * @throws {CertificateVerificationError}
+   */
+  public static async create(
```
krpeacock, line 108: For constructors, we prefer to use a single options object instead of positional arguments. Do you mind pulling these out into a new type, and then documenting the object's properties in the JSDOC comment? See https://github.com/dfinity/agent-js/blob/main/packages/auth-client/src/index.ts#L212 as a reference

krpeacock (changes requested):

robin-kunzler (approved):
This looks good to me, thanks a lot for implementing the range check @oggy-dfin!

Please find a few comments below.

packages/agent/src/certificate.ts
```diff
@@ -99,59 +98,110 @@ function isBufferEqual(a: ArrayBuffer, b: ArrayBuffer): boolean {
...
+   * The bytes encoding the certificate to be verified
+   */
+  certificate: ArrayBuffer;
+  /**
+   * The root key against which to verify the certificate
+   * (normally, the root key of the IC main network)
+   */
+  rootKey: ArrayBuffer;
+  /**
+   * The effective canister ID of the request that generated the certificate.
```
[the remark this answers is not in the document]
oggy-dfin: The canister doesn't necessarily generate the certificate - the result can be just the call to `read_state` (for example, to read the status of a request). But you're right, it can also be generated by the canister. I think we can say effective canister ID/signing canister ID, depending on the context. I'll formulate something along those lines.

packages/agent/src/certificate.test.ts
```diff
@@ -124,3 +133,67 @@ test('lookup', () => {
...
+    }),
+  ).resolves.not.toThrow();
+});
+
+function fail(reason) {
+  throw new Error(reason);
+}
+
+test('delegation check fails for canisters outside of the subnet range', async () => {
+  // Use a different principal than the happy path, which isn't in the delegation ranges.
```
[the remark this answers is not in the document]
frederikrothenberger: You can actually decode this (it's cbor, https://cbor.me comes in handy).

So the first level of decoding reveals
```
"delegation": {"subnet_id": h'D77B2A2F7199B9A8AEC93FE6FB588661358CF12223E9A3AF7B4EBAC402', "certificate": h'D9D9F7A26474726565830182045820AE023F28C3B9D966C8FB09F9ED755C828AADB5152E00AAF700B18C9C067294B483018302467375626E6574830182045820E83BB025F6574C8F31233DC0FE289FF546DFA1E49BD6116DD6E8896D90A4946E830182045820E782619092D69D5BEBF0924138BD4116B0156B5A95E25C358EA8CF7E7161A661830183018204582062513FA926C9A9EF803AC284D620F303189588E1D3904349AB63B6470856FC4883018204582060E9A344CED2C9C4A96A0197FD585F2D259DBD193E4EADA56239CAC26087F9C58302581DD77B2A2F7199B9A8AEC93FE6FB588661358CF12223E9A3AF7B4EBAC402830183024F63616E69737465725F72616E6765738203581BD9D9F781824A000000000020000001014A00000000002FFFFF010183024A7075626C69635F6B657982035885308182301D060D2B0601040182DC7C0503010201060C2B0601040182DC7C050302010361009933E1F89E8A3C4D7FDCCCDBD518089E2BD4D8180A261F18D9C247A52768EBCE98DC7328A39814A8F911086A1DD50CBE015E2A53B7BF78B55288893DAA15C346640E8831D72A12BDEDD979D28470C34823B8D1C3F4795D9C3984A247132E94FE82045820996F17BB926BE3315745DEA7282005A793B58E76AFEB5D43D1A28CE29D2D158583024474696D6582034995B8AAC0E4EDA2EA16697369676E61747572655830ACE9FCDD9BC977E05D6328F889DC4E7C99114C737A494653CB27A1F55C06F4555E0F160980AF5EAD098ACC195010B2F7'}
```
The included delegation cert then specifies for the subnet (in addition to it's public key) the canister_ranges `D9D9F781824A000000000020000001014A00000000002FFFFF0101` (which is again cbor. Decoded this results in the following ranges:

`[(CanisterId(jrlun-jiaaa-aaaab-aaaaa-cai), CanisterId(v2nog-2aaaa-aaaab-p777q-cai))]`

Which are the following principals:
`Principal(PrincipalInner { len: 10, bytes: [0, 0, 0, 0, 0, 32, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }), Principal(PrincipalInner { len: 10, bytes: [0, 0, 0, 0, 0, 47, 255, 255, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] })
`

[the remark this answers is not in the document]
oggy-dfin: Thanks @frederikrothenberger ! 

I now realized that the interface spec also provides handy tools for decoding the text representation (which indeed doesn't preserve the lexicographic ordering, due to the use of crc-32).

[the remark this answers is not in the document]
oggy-dfin: Added the checks now, switched to a hex representation of the principals as that makes the ranges clearly visible

packages/agent/src/certificate.test.ts
```diff
@@ -124,3 +133,67 @@ test('lookup', () => {
...
+  ).resolves.not.toThrow();
+});
+
+type FakeCert = {
+  tree: Cert.HashTree;
+  signature: ArrayBuffer;
+  delegation?: { subnet_id: ArrayBuffer; certificate: ArrayBuffer };
+};
+
+test('certificate verification fails for an invalid signature', async () => {
```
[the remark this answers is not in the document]
oggy-dfin: It makes total sense, certification is important. Created [a ticket](https://dfinity.atlassian.net/browse/SDK-571) now

krpeacock (approved):

packages/agent/src/certificate.test.ts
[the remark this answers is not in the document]
robin-kunzler: Great, thanks for addressing this @oggy-dfin !

packages/agent/src/certificate.test.ts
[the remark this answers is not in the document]
robin-kunzler: Perfect, thanks for addressing the ticket.

PR #584, opened 2022-06-28
feat: reuse signed request when reading state

lmuntaner:
# Description

Enable the `readState` of the Agent to reuse the request.

## Motivation

Signing with the hardware wallet requires the user to accept from the device. Reusing the signed request will improve the user experience for hardware wallet users.

# How Has This Been Tested?

I linked a branch of nns-dapp to the changes in this PR and validated that everything worked, not just the new feature.

# Checklist:

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -352,12 +352,11 @@ export class HttpAgent implements Agent {
     return cbor.decode(await response.arrayBuffer());
   }
 
-  public async readState(
-    canisterId: Principal | string,
+  public async createReadStateRequest(
     fields: ReadStateOptions,
     identity?: Identity | Promise<Identity>,
-  ): Promise<ReadStateResponse> {
-    const canister = typeof canisterId === 'string' ? Principal.fromText(canisterId) : canisterId;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ): Promise<any> {
     const id = await (identity !== undefined ? await identity : await this._identity);
     if (!id) {
       throw new IdentityInvalidError(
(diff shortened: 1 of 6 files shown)


lmuntaner:
@krpeacock could you review these changes?

I would love to remove the `any` type. Yet, the transform returns `unknwon` so not sure how I can go around that. Do you have any idea?

packages/agent/src/agent/http/index.ts
```diff
@@ -386,8 +385,19 @@ export class HttpAgent implements Agent {
     });
 
     // Apply transform for identity.
-    transformedRequest = await id?.transformRequest(transformedRequest);
+    return id?.transformRequest(transformedRequest);
```
krpeacock, line 388: @lmuntaner is removing this `await` intentional?

lmuntaner: No, just because I am returning the value, I don't think it's needed.

If we had a try/catch then yes, but since we don't, then I don't think so.

krpeacock:
@lmuntaner I think the approach here is good, and I see the usefulness for hardware wallets where each request requires another prompt. 

It's unfortunate to expand the use of `any`, but you're using it in the same way it's already being used, so I'm not going to make you refactor the script to get this feature in. 

I'm ready to approve, if we add a new e2e test and document the new feature in the changelog

lmuntaner:
Yes, I agree the use of `any` is very ugly. I haven't found a better way unfortunately.

The test and docs sound good, when I have some time I will get to it. Thanks.

lmuntaner:
@krpeacock ready for another review

I added an e2e tests to read_state to make sure it also works when the request is passed. I also added a few points to the changelog.

Let me know what you think.

lmuntaner:
@krpeacock is the e2e test and the docs that I added what you needed??

krpeacock (approved):
Looks good, thanks for the work on this one!

PR #598, opened 2022-07-25
fix: IDL edge case involving Math.log2 function on BigInt bit count

rvanasa:
# Description

Fixes a bug where `Math.log2` is sometimes called on a BigInt value in `@dfinity/candid` (possibly a regression from #575), causing an unexpected error in Motoko Playground. 

This PR adds a custom `ilog2` implementation ([based on this StackOverflow post](https://stackoverflow.com/questions/55355184/optimized-integer-logarithm-base2-for-bigint)) which accounts for BigInt values while providing clearer error messages when receiving invalid input. 

# How Has This Been Tested?

- Full repository test suite
- Additional test cases for `ilog2` function

# Checklist:

- [ ] I have edited the CHANGELOG accordingly. (N/A, regression)
- [ ] I have made corresponding changes to the documentation. (N/A)

### DIFF
--- a/packages/candid/src/idl.ts
+++ b/packages/candid/src/idl.ts
@@ -15,6 +15,7 @@ import {
   writeIntLE,
   writeUIntLE,
 } from './utils/leb128';
+import { ilog2 } from './utils/ilog2';
 
 // tslint:disable:max-line-length
 /**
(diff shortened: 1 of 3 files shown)


packages/candid/src/idl.ts
```diff
@@ -647,7 +648,7 @@ export class FixedIntClass extends PrimitiveType<bigint | number> {
   }
 
   public encodeType() {
-    const offset = Math.log2(this._bits) - 3;
+    const offset = ilog2(this._bits) - 3;
```
chenyan-dfinity, line 651: Is this the right place for the fix? `this._bits` can only be `8,16,32,64`. I would expect the problem comes from taking log2 of the actual number, not `_bits`?

rvanasa:
Closing in favor of a different solution.

PR #605, opened 2022-08-04
feat: auth-client stores identities in indexeddb

krpeacock:
# Description

AuthClient now uses IndexedDb by default, as localStorage is less secure.
Also offers a generic Indexed Db keyval store, IdbKeyVal

I decided to leave out encryption for this PR to reduce overhead - could be added at a later date https://gist.github.com/krpeacock/9203f1c6bbd8c55f2eef5e001223edf2

Fixes SDK-313

# How Has This Been Tested?

Unit test coverage

# Checklist:

### DIFF
--- a/packages/auth-client/src/db.ts
+++ b/packages/auth-client/src/db.ts
@@ -1,0 +1,45 @@
+import { openDB, IDBPDatabase } from 'idb';
+
+export type Database = Promise<IDBPDatabase<unknown>>;
+
+export const openDbStore = async () =>
+  await openDB('auth-client-db', 1, {
+    upgrade: database => {
+      if (database.objectStoreNames.contains('ic-idp')) {
+        database.clear('ic-idp');
+      }
+      database.createObjectStore('ic-idp');
+    },
+  });
+
+/**
+ *
+ * @param db database
+ * @param key string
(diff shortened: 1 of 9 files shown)


packages/auth-client/src/db.ts
```diff
@@ -0,0 +1,110 @@
...
+const AUTH_DB_NAME = 'auth-client-db';
+const OBJECT_STORE_NAME = 'ic-keyval';
+
+const _openDbStore = async (
+  dbName = AUTH_DB_NAME,
+  storeName = OBJECT_STORE_NAME,
+  version: number,
+) => {
+  // Clear legacy stored delegations
+  if (localStorage && localStorage.getItem(KEY_STORAGE_DELEGATION)) {
```
ericswanson-dfinity, line 15: If this changes the place where identities are stored, is there any migration of identities from the previous place to the new one?  Or is there a reason why that would not be necessary?

krpeacock: Identities are temporary delegation identities, and get cleared on `logout`. Resetting after a package upgrade is a fairly safe interaction, and will only require a new II login

ericswanson-dfinity (approved):

PR #614, opened 2022-08-23
feat: principal backwards compatibility

krpeacock:
# Description

Introduces a PrincipalBase abstract class that can be used to ensure compatibility with agent-js 0.9.3

# Checklist:


packages/principal/src/index.test.ts
```diff
@@ -42,3 +43,35 @@ describe('Principal', () => {
...
+    const anonymousLegacy = Principal.anonymous();
+    expect(anonymous.toText()).toMatch(anonymousLegacy.toText());
+
+    const principal1 = Principal.fromText('ryjl3-tyaaa-aaaaa-aaaba-cai');
+    const principal2 = PrincipalLegacy.fromText('ivg37-qiaaa-aaaab-aaaga-cai');
+
+    expect(principal1.compareTo(principal2)).toBe('lt');
+  });
+  test('assignment to PrincipalBase', () => {
+    const current = Principal.anonymous() as PrincipalBase;
```
ericswanson-dfinity, line 73: What is the purpose of `current` in this test?  (Wondering about the warnings)

krpeacock: It's to verify that the assignment - `as PrincipalBase` doesn't throw TypeScript errors. If the signature was incompatible, it would mean that the abstract class is no longer compatible with the current version

krpeacock:
The one thing I'm unsure of is the naming strategy - I've gone with `PrincipalBase`, but I may prefer `AbstractPrincipal` as the convention. There is no rush in merging this while I gather feedback and think about it

docs/generated/changelog.html
```diff
@@ -10,6 +10,17 @@
     <h1>Agent-JS Changelog</h1>
 
     <section>
+      <h2>Version 1.0.0</h2>
+      <ul>
+        <li>Backwards compatibility with 0.9.3 for @dfinity/principal</li>
+        <ul>
+          <li>
+            If you have a TypeScript interface that uses a Principal, we recommend updating and
+            using the abstract class PrincipalBase as the signature. It will be compatible across
```
ericswanson-dfinity, line 19:
```suggestion
            using the abstract class PrincipalLike as the signature. It will be compatible across
```

ericswanson-dfinity (approved):

PR #617, opened 2022-08-31
fix: return after resolve to avoid idb to be recreated

peterpeterparker:
# Description

While I did not actually face any issue related to this PR, based on the code, a `return` seems to be missing after promise resolution when the instance of the idb connection is requested within `IdbStorage` - i.e. even though the connection exists, it might be recreated anyway.


peterpeterparker:
Done, added an entry in changelog.html for this PR and your last week's [fix](https://github.com/dfinity/agent-js/pull/620)

krpeacock (approved):

krpeacock (approved):

PR #623, opened 2022-09-01
feat: expose an agent syncTime method

krpeacock:
# Description

Adds a `syncTime` method to HttpAgent, allowing the agent to set an expiry within the
          acceptable bounds for a replica even if the device clock is out of sync

Fixes # (issue)

# How Has This Been Tested?

New unit test suite

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -418,6 +427,25 @@ export class HttpAgent implements Agent {
...
 
+  /**
+   * Allows agent to sync its time with the network. Can be called during intialization or mid-lifecycle if the device's clock has drifted away from the network time. This is necessary to set the Expiry for a request
+   * @param {PrincipalLike} canisterId - Pass a canister ID if you need to sync the time with a particular replica. Uses the management canister by default
+   */
+  public async syncTime(canisterId?: PrincipalLike): Promise<void> {
+    const CanisterStatus = await import('../../canisterStatus');
+    const callTime = Date.now();
+    const status = await CanisterStatus.request({
+      canisterId: canisterId ?? Principal.managementCanister(),
```
chenyan-dfinity, line 438: I can be wrong, but you cannot get the effective canister id from managementCanister, so I guess boundary node will fail to forward your request to IC?

krpeacock: You're right! I just tested it against the IC. Is there any appropriate "default" principal that will always resolve?

chenyan-dfinity: I think you can use a well-known canister id, for example, the ledger or NNS canister id.

PR #624, opened 2022-09-04
fix: Cbor.encode(BigInt(n)) possible failure

neeboo:
Fix `Cbor.encode(BigInt(n))` possible failure

# Description

When use `Cbor.encode`, if passed value is BigInt, first will run the BigInt encoder and transform BigInt value to be a hex string. However this `BigInt(n).toString(16)` would probrably generate a hex string length not to be even, which won't pass the RegTest in the `fromHex` function. 

```typescript
const hexRe = new RegExp(/^([0-9A-F]{2})*$/i);
```

Fixes # (issue)

Added a zero-padding while passing hex string length is not even.

```typescript
export function fromHex(hex: string): ArrayBuffer {
  //FIXME: always padding hex string length to be even
  // then run the reg test
  if (hex.length % 2 !== 0) {
    hex = '0' + hex;
  }

  if (!hexRe.test(hex)) {
    throw new Error('Invalid hexadecimal string.');
  }
  const buffer = [...hex]
    .reduce((acc, curr, i) => {
      // tslint:disable-next-line:no-bitwise
      acc[(i / 2) | 0] = (acc[(i / 2) | 0] || '') + curr;
      return acc;
    }, [] as string[])
    .map(x => Number.parseInt(x, 16));

  return new Uint8Array(buffer).buffer;
}


```

# How Has This Been Tested?

test cases are in `cbor.tests.ts`

# Checklist:

### DIFF
--- a/packages/agent/src/cbor.test.ts
+++ b/packages/agent/src/cbor.test.ts
@@ -54,6 +54,82 @@ test('empty canister ID', () => {
   expect(Principal.fromUint8Array(outputA as any).toText()).toBe('aaaaa-aa');
 });
 
+describe('cbor encode/decode bigint', () => {
+  test('encode decode 0 to 1_000_000', () => {
+    for (let i = 0; i < 1_000_000; i += 1) {
+      let buffer;
+
+      try {
+        buffer = encode(BigInt(i));
+      } catch (error) {
+        console.log(`error encoding on ${i.toString()}`);
+        expect((error as Error).message).toBeTruthy();
+      }
+
+      try {
+        const decoded = decode<bigint>(buffer);
+        expect(decoded).toBe(BigInt(i));
+      } catch (error) {
+        console.log(`error decoding on ${i.toString()}`);
+        expect((error as Error).message).toBeTruthy();
+      }
+      // if (i % 100000 === 0) {
+      //   console.log(`${new Date(Date.now()).toLocaleTimeString()} with ${i.toString()}`);
+      // }
+    }
+  });
+  test('encode decode 1_000_000n to 1_000_000_000_000n', () => {
+    for (let i = 1_000_000; i < 1_000_000_000_000; i += 1_000_000) {
+      let buffer;
+
+      const random_num = Math.floor(Math.random() * Math.pow(10, i.toString().length - 1));
+
+      const actual = BigInt(i + random_num);
+      try {
+        buffer = encode(actual);
+      } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/agent/src/cbor.test.ts
```diff
@@ -54,6 +54,85 @@ test('empty canister ID', () => {
...
+        expect((error as Error).message).toBeTruthy();
+      }
+      // // if we log the time
+      // if (i % 1_000_000_000_00 === 0) {
+      //   console.log(`${new Date(Date.now()).toLocaleTimeString()} with ${actual.toString()}`);
+      // }
+    }
+  });
+  test('encode decode 1_000_000_000_000n to 1_000_000_000_000_000_000n', () => {
+    for (let i = 1_000_000_000_000n; i < 1_000_000_000_000_000_000n; i += 1_000_000_000_000n) {
```
krpeacock, line 110:
```suggestion
    for (let i = BigInt(1_000_000_000_000); i < BigInt(1_000_000_000_000_000_000); i += BigInt(1_000_000_000_000)) {
```

krpeacock: to make typescript happy

packages/agent/src/utils/buffer.ts
```diff
@@ -27,6 +27,12 @@ const hexRe = new RegExp(/^([0-9A-F]{2})*$/i);
  * @param hex The hexadecimal string to use.
  */
 export function fromHex(hex: string): ArrayBuffer {
+  //FIXME: always padding hex string length to be even
```
krpeacock, line 30: What's the rationale for this? Does decoding break for odd-length hex strings?

packages/agent/src/cbor.test.ts
```diff
@@ -54,6 +54,85 @@ test('empty canister ID', () => {
...
+      try {
+        const decoded = decode<bigint>(buffer);
+        expect(decoded).toBe(BigInt(i));
+      } catch (error) {
+        console.log(`error decoding on ${i.toString()}`);
+        expect((error as Error).message).toBeTruthy();
+      }
+      // // if we log the time
+      // if (i % 100000 === 0) {
+      //   console.log(`${new Date(Date.now()).toLocaleTimeString()} with ${i.toString()}`);
```
krpeacock, line 78: Is there a reason these comments are left in?

neeboo: Can delete those

PR #627, opened 2022-09-07
fix: optional fields not populated if wire type has additional fields

hpeebles:
# Description

If the wire type contains more fields than the output type, then (dependent upon the field hash ordering) optional fields in the output type can be skipped even if they match with types on the wire.

# How Has This Been Tested?

I have added a unit test which covers this case and all the existing units tests also pass.

# Checklist:


hpeebles:
This PR makes the assumption that fields are always ordered by their hash values within the type definition. Is that guaranteed by the candid spec?

chenyan-dfinity (approved):
Thanks for the fix! The field is sorted as guaranteed by the spec.

nomeata:
Is this something that the official candid test at https://github.com/dfinity/candid/tree/master/test suite covers?

If yes, it may be worth investing the effort to run agent-js against it somehow?

If not, please extend that test suite maybe?

chenyan-dfinity:
I will check. The JS implementation isn't doing the subtype checking yet, so we didn't run it through the test suite.

PR #630, opened 2022-09-09
feat: Add fetchCandid() function to @dfinity/agent

rvanasa:
# Description

This PR adds a top-level function to abstract away the implementation details of the hidden `__get_candid_interface_tmp_hack` method in each canister. 

At the moment, several projects (notably the [IC Dashboard](https://dashboard.internetcomputer.org/) and [Candid UI](https://a4gq6-oaaaa-aaaab-qaa4q-cai.raw.ic0.app/)) directly rely on this unstable `_tmp_hack` method. This `fetchCandid()` function can be used as a replacement in both of these applications. 

# How Has This Been Tested?

- Added a new test case: `agent/src/fetch_candid.test.ts`
- Works as expected in a browser environment. 

# Checklist:

- [x] I have made corresponding changes to the documentation. (JSDoc)

### DIFF
--- a/packages/agent/src/fetch_candid.ts
+++ b/packages/agent/src/fetch_candid.ts
@@ -1,0 +1,21 @@
+import { Actor, ActorSubclass, Agent } from '.';
+import { IDL } from '@dfinity/candid';
+
+// Common interface for the hidden `__get_candid_interface_tmp_hack` actor method
+const tmpHackInterface: IDL.InterfaceFactory = ({ IDL }) =>
+  IDL.Service({
+    __get_candid_interface_tmp_hack: IDL.Func([], [IDL.Text], ['query']),
+  });
+
+/**
+ * Retrieves the Candid interface for the specified canister..
+ *
+ * @param agent The agent to use for the request (usually an `HttpAgent`)
+ * @param canisterId A string corresponding to the canister ID
+ * @returns Candid source code
+ */
+export async function fetchCandid(agent: Agent, canisterId: string): Promise<string> {
+  const actor: ActorSubclass = Actor.createActor(tmpHackInterface, { agent, canisterId });
+  const candidSource = (await actor.__get_candid_interface_tmp_hack()) as string;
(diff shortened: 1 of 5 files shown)


packages/agent/src/fetch_candid.ts
```diff
@@ -0,0 +1,21 @@
...
+  });
+
+/**
+ * Retrieves the Candid interface for the specified canister..
+ *
+ * @param agent The agent to use for the request (usually an `HttpAgent`)
+ * @param canisterId A string corresponding to the canister ID
+ * @returns Candid source code
+ */
+export async function fetchCandid(agent: Agent, canisterId: string): Promise<string> {
```
chenyan-dfinity, line 17: We need to use this logic: https://github.com/dfinity/candid/blob/master/tools/ui/src/candid.ts#L57

Fetch from metadata first, if it fails, fetch from `_tmp_hack`. All canisters built from recent dfx releases should have the metadata already. But we can never drop `_tmp_hack` because some canisters never upgrade.

rvanasa: Makes sense. This function now uses the same logic as in the Candid UI.

chenyan-dfinity (approved):

krpeacock:
I think this is more intuitive than CanisterStatus by default! One suggestion - since `candid` is public metadata, why not move `agent` to the second position and make it optional, using an anonymous `HttpAgent` by default?

rvanasa:
Sounds good; updated to use the same anonymous HttpAgent logic from both Candid UI and Motoko Playground. 

I also added `HttpAgent::isLocal()` to simplify the relatively common use case of fetching the root key for local environments. Let me know if there's a better way of approaching this from within the `agent-js` codebase.

krpeacock:
We've avoided the concept of `local` or `mainnet` for the sake of handling testnets more naturally for NNS dapps, but I don't think I mind introducing this pattern now - it's useful for most use cases

PR #657, opened 2022-11-22
docs: Small edit in JS Doc comment in Auth-Client Package

Web3NL:
After switch to idbStorage, comment was left unedited


krpeacock:
I think the CI broke with the latest SDK update. Can you merge the latest from `main`?

krpeacock (approved):

PR #681, opened 2023-01-24
feat: introduces new package @dfinity/core

krpeacock:
# Description

Since it was becoming a pain to install the individual packages, since most of them were already necessary for the auth-client anyway, this PR introduces `@dfinity/core`, a bundle of 

- @dfinity/agent
- @dfinity/candid
- @dfinity/principal
- @dfinity/identity
- @dfinity/utils

So you can rapidly get to work building instead of maintaining versions. The individual packages will still be maintained and released separately for projects that have specific needs and want to reduce bloat.

# Checklist:


peterpeterparker:
Not sure `@dfinity/utils` should be embedded. I like the idea that we provide it to more devs, because useful, but most do not use it currently. They don't necessecary have to, on the contrary of the other libs here that are almost mandatory to implement anything on the IC.

lmuntaner:
> Not sure `@dfinity/utils` should be embedded. I like the idea that we provide it to more devs, because useful, but most do not use it currently. They don't necessecary have to, on the contrary of the other libs here that are almost mandatory to implement anything on the IC...

I agree with @peterpeterparker, I'm not sure that utils should be part of the core. It's very specific to NNS, except for the helpers `defaultAgent`, `createAgent` and `createServices`.

Not a strong opinion either, but I also agree on moving it here if we do so.

krpeacock:
My view is that there should be a base set of utils, that even `candid`, `principal`, and `agent` depend on. Of the items in utils, I imagine the following are all broadly useful to people:

- [defaultAgent](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-defaultagent)
- [createAgent](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-createagent)
- [uint8ArrayToBigInt](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-uint8arraytobigint)
- [bigIntToUint8Array](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-biginttouint8array)
- [numberToUint8Array](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-numbertouint8array)
- [arrayBufferToUint8Array](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-arraybuffertouint8array)
- [uint8ArrayToArrayOfNumber](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-uint8arraytoarrayofnumber)
- [arrayOfNumberToUint8Array](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-arrayofnumbertouint8array)
- [assertNonNullish](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-assertnonnullish)
- [toNullable](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-tonullable)
- [fromNullable](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-fromnullable)
- [fromDefinedNullable](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-fromdefinednullable)
- [principalToSubAccount](https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-principaltosubaccount)

I also would support moving the buffer and leb utility methods out of `candid` and into `utils`

lmuntaner:
@krpeacock @peterpeterparker I came across this again.

Should we continue with this? Did something change?

krpeacock:
@lmuntaner After doing some exploration of other packages, I decided that it would be a better approach to fold all of the packages into `@dfinity/agent`, and to have all the exports be from subdirectories instead of the root. Basically the same structure that https://www.npmjs.com/package/@noble/curves uses.

The refactor was more than I could pull off in a cooldown week (I did try), so I'd need to prioritize it as a feature if I wanted to get this through

lmuntaner:
@krpeacock that makes sense.

Does `@dfinity/agent` use all those utils? I'm not so sure all of them fit so well there though. For example:
assertNonNullish
toNullable
fromNullable
fromDefinedNullable

Those seem from a different domain than agent. Maybe inside `@dfinity/candid`?

PR #689, opened 2023-02-15
feat: auth client keyType config option

krpeacock:
# Description

Rem.codes pointed out that the ECDSA keys weren't working in a localstorage implementation. This feature adds the option to choose between ECDSA and ED25519 keys for the auth client base key.

It will also warn developers if they are using a custom storage implementation with "ECDSA" that it may not work correctly, and that they may want to consider using ED25519.

# How Has This Been Tested?

new unit test

# Checklist:


packages/auth-client/src/index.ts
```diff
@@ -203,10 +220,13 @@ export class AuthClient {
           const fallbackLocalStorage = new LocalStorage();
           const localChain = await fallbackLocalStorage.get(KEY_STORAGE_DELEGATION);
           const localKey = await fallbackLocalStorage.get(KEY_STORAGE_KEY);
-          if (localChain && localKey) {
+          // not relevant for Ed25519
+          if (localChain && localKey && keyType === 'ECDSA') {
             console.log('Discovered an identity stored in localstorage. Migrating to IndexedDB');
             await storage.set(KEY_STORAGE_DELEGATION, localChain);
+            localKey; //?
```
dfx-json, line 227: what does this line do?

packages/auth-client/src/index.ts
```diff
@@ -219,7 +239,11 @@ export class AuthClient {
       if (maybeIdentityStorage) {
         try {
           if (typeof maybeIdentityStorage === 'object') {
-            key = await ECDSAKeyIdentity.fromKeyPair(maybeIdentityStorage);
+            if (keyType === 'Ed25519' && typeof maybeIdentityStorage === 'string') {
```
dfx-json, line 242: I wonder if strings like `Ed25519` and `ECDSA` would be better off defined as enums or consts.

dfx-json (approved):
Left a couple comments. but otherwise, LGTM

packages/auth-client/src/index.ts
krpeacock: oh, that's a comment. I'll remove it

packages/auth-client/src/index.ts
krpeacock: I can move it to a const

krpeacock: although, that doesn't play well with the actual parameter

PR #694, opened 2023-02-25
fix: reverts use of Headers for improved node compatibility

### LINKED ISSUE #685: Headers is not used in nodejs env.
https://github.com/dfinity/agent-js/blob/aa14db37898852cb7293aeda67ed09dc9ea61a91/packages/agent/src/agent/http/index.ts#L386

Problem:
`Headers` of browser is not used in nodejs environment. If we use agent in nodejs, it will be giving error like following:

```bash
ReferenceError: Headers is not defined
    at HttpAgent.call 
...
```

Proposal:
Consider to use `cross-fetch` and switch Headers between different env.
```typescript
import { fetch, Headers } from 'cross-fetch';

if (!globalThis.fetch) {
  (globalThis as any).fetch = fetch;
  (globalThis as any).Headers = Headers;
}

if (!global.fetch) {
  (global as any).fetch = fetch;
  (global as any).Headers = Headers;
}


 new Headers()
...
```

krpeacock:
# Description

Due to the use of Headers inside of Agent-JS, it introduced a nodejs compatibility issue. We can simply swap back with no impact on the functionality

Fixes #685 

# How Has This Been Tested?

Unit tests

# Checklist:

### DIFF
--- a/docs/generated/changelog.html
+++ b/docs/generated/changelog.html
@@ -13,6 +13,11 @@
       <h2>Version x.x.x</h2>
       <ul>
         <li>
+          Do not use `Headers` struct during init of HttpAgent for Node compatibility. Note: still
+          supports use of Headers in application code
+        </li>
+        <li></li>
+        <li>
           fix: finish all tasks before calling onSuccess auth callback in @dfinity/auth-client
         </li>
       </ul>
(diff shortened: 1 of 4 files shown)


ericswanson-dfinity (commented):
Is there a note in the changelog to be reverted or amended?

krpeacock:
@ericswanson-dfinity I don't think it's necessary - I'm keeping type support for someone to pass a `Headers` struct if they want to, but no longer using it by default in the implementation. Simply adding a note to the changelog should be adequate

ericswanson-dfinity (approved):

PR #698, opened 2023-03-06
feat(agent): adds transaction hashes

DaneWiid1:
# Description
Adds a feature to the agent package to generate transaction hashes. 

Generates transaction hashes to are identical to Rosetta transaction hashes without needing to run a Rosetta node purely for this purpose.

This pr includes a new CBOR dependency https://www.npmjs.com/package/cbor which supports collapsing BigInt. If preferred, the code to collapse big integers could be moved into agent-js to remove the additional dependency. 

Fixes # **having to run a rosetta node to generate a transaction hash**

# How Has This Been Tested?

Jest unit tests have been added to cover the feature.

# Checklist:

**comment:** no change log found
**comment:** Above the granularity of the readme. No other documentation found.


krpeacock:
The interface looks clean! I think this may be a little too specialized for the core agent package, since we want to keep that as lean as possible since it's bundled in every web UI across the ecosystem, and not all applications need to be concerned with transactions. 

@nathanosdev @peterpeterparker what do you think about where this should live?

nathanosdev:
I agree this is a very clean contribution! Thank you @DaneWiid1 

Moving this into a separate package would be better in my opinion, but I don't have a strong opinion about where that package should live.

The agent-js repo houses the agent but also authentication libraries, so it has more of a JS SDK monorepo feel to me now and I think adding a sibling package for transactions wouldn't be too big of a jump.

DaneWiid1:
Hi @krpeacock, @nathanosdev 

Thank you for taking the time to review my pull request. 
I have read through the terms of your CLA, and unfortunately I cannot agree with it.
We will reach out through other channels to see if we can resolve this.  

In the meantime hopefully this can provide a useful reference for anyone implementing this functionality in their project.

krpeacock:
Closing this PR, as Dane has decided to withdraw

PR #705, opened 2023-03-30
feat: Retry failed `read_state` requests

hpeebles:
# Description

When making an `update` request to the IC, the agent first submits a `call` request, then it polls `read_state` until the result is ready. If the `call` request fails it would be retried, but that same retry logic wasn't being applied to the `read_state` requests, so if any request to `read_state` failed, the caller would receive an error even though their request may still complete successfully.

This PR wraps calls to `read_state` in the same retry pattern which is already in use on `call`, `query` and `status` requests.

# How Has This Been Tested?

It hasn't yet, I'm about to test it on OpenChat locally.

# Checklist:

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -466,13 +466,12 @@ export class HttpAgent implements Agent {
     const transformedRequest = request ?? (await this.createReadStateRequest(fields, identity));
     const body = cbor.encode(transformedRequest.body);
 
-    const response = await this._fetch(
-      '' + new URL(`/api/v2/canister/${canister}/read_state`, this._host),
-      {
+    const response = await this._requestAndRetry(() =>
+      this._fetch('' + new URL(`/api/v2/canister/${canister}/read_state`, this._host), {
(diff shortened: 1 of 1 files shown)


krpeacock:
I'd like to add some unit tests to this before merging - I can get to it next week if you don't have the time

packages/agent/src/agent/http/index.ts
```diff
@@ -466,13 +466,12 @@ export class HttpAgent implements Agent {
     const transformedRequest = request ?? (await this.createReadStateRequest(fields, identity));
     const body = cbor.encode(transformedRequest.body);
 
-    const response = await this._fetch(
-      '' + new URL(`/api/v2/canister/${canister}/read_state`, this._host),
-      {
+    const response = await this._requestAndRetry(() =>
```
krpeacock, line 469:
```suggestion
    // TODO - https://dfinity.atlassian.net/browse/SDK-1092
    const response = await this._requestAndRetry(() =>
```

krpeacock (approved):

PR #714, opened 2023-04-22
fix: finish all tasks before calling onSuccess callback

nathanosdev:
# Description

The `onSuccess` callback was called before writing the delegation chain to storage. Since the writing of the chain to storage is asynchronous, an app that performs some full page navigation or a page refresh in their `onSuccess` callback _might_ interrupt the operation.

# How Has This Been Tested?

This was causing flakiness in some e2e tests that I was working on for a dapp. Roughly 20% of the runs were failing. Making this change removes the flakiness and the tests pass now 100% of the time.

# Checklist:

### DIFF
--- a/docs/generated/changelog.html
+++ b/docs/generated/changelog.html
@@ -12,7 +12,9 @@
     <section>
       <h2>Version x.x.x</h2>
       <ul>
-        <li></li>
+        <li>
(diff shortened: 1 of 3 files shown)


krpeacock (approved):
Nice - this is easier to read, as well!

PR #730, opened 2023-06-27
feat: support composite_query in candid

ulan:
This change adds a new query type: `composite_query` in idl.ts.

It corresponds to the following changes in the upstream candid:
- https://github.com/dfinity/candid/pull/420 (spec change)
- https://github.com/dfinity/candid/pull/435 (rust change)


ulan:
> I have made corresponding changes to the documentation.

I wasn't sure if there are any docs that needs updating here.

THLO (approved):

PR #734, opened 2023-07-05
fix: handle new update call errors (IC-1462)

rvanasa:
# Description

Handles the new update call error messages introduced in [IC-1462](https://github.com/dfinity/interface-spec/pull/143). 

Corresponds to https://github.com/dfinity/agent-rs/pull/422. 

# How Has This Been Tested?

This PR includes a unit test for the new update call behavior. 

I also ran a local sanity check using the setup described in [this forum post](https://forum.dfinity.org/t/not-calling-accept-message-in-inspect-message-not-rejecting-immediately-in-dfx-0-14-2-beta-2/21105), replacing Candid UI with a simple webapp pointing to a local `agent-js` installation. 

# Checklist:

- [x] I have made corresponding changes to the documentation. (N/A)

[IC-1462]: https://dfinity.atlassian.net/browse/IC-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/agent/src/actor.ts
+++ b/packages/agent/src/actor.ts
@@ -58,13 +58,13 @@ export class UpdateCallRejectedError extends ActorCallError {
   ) {
     super(canisterId, methodName, 'update', {
       'Request ID': toHex(requestId),
-      ...(response.body?.error_code
-        ? {
-            'Error code': response.body.error_code,
-          }
-        : {}),
       ...(response.body
         ? {
+            ...(response.body.error_code
+              ? {
+                  'Error code': response.body.error_code,
+                }
+              : {}),
             'Reject code': String(response.body.reject_code),
             'Reject message': response.body.reject_message,
           }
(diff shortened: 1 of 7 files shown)


dfx-json (approved):
🙏🏼

packages/agent/src/actor.ts
```diff
@@ -364,7 +376,7 @@ function _createActorMethod(
         effectiveCanisterId: ecid,
       });
 
-      if (!response.ok) {
+      if (!response.ok || response.body /* IC-1462 */) {
```
ericswanson-dfinity, line 379: Would you please describe a little more here why we would treat "the response has a body" as "Update call rejected" ?  I'm reading the description for IC-1462 and don't understand.

rvanasa: From the [spec](https://github.com/dfinity/interface-spec/blob/master/spec/index.md#request-call-http-call):

image

I don't fully understand the reasoning for this change, but this is indeed the correct behavior for the latest `dfx` release.

ericswanson-dfinity (approved):

PR #736, opened 2023-07-07
feat: expose boundary node http headers to calls

keplervital:
# Description

This PR extens the `query` and `call` method responses to include the boundary node http headers. Moreover, it introduces a new method to the actor creation that enriches the provide type to include the http details of call.

# How Has This Been Tested?

- http calls responses are tested with snapshots
- `requestAndRetry` logic is validated by using a custom error type `AgentHTTPResponseError `
- actor interface is tested by mocking the pollForResponse and testing that both `createActor` and `createActorWithHttpDetails` return the same decoded response and that the later also includes the expected `httpDetails`
- Linked locally with the service worker and manually validated the integration

# Checklist:

### DIFF
--- a/docs/generated/changelog.html
+++ b/docs/generated/changelog.html
@@ -32,9 +32,11 @@
           in the type table
         </li>
         <li>
-          feat: support composite_query in candid
-          fix: fix a bug in decoding service types, when function types come after the service type in the type table
+          feat: support composite_query in candid fix: fix a bug in decoding service types, when
+          function types come after the service type in the type table
         </li>
+        <li>feat: include boundary node http details to query and update calls</li>
+        <li>feat: adds method for actor creation that includes boundary node http details</li>
       </ul>
       <h2>Version 0.15.7</h2>
       <ul>
--- a/docs/generated/changelog.html
+++ b/docs/generated/changelog.html
@@ -31,9 +31,10 @@
           fix: fix a bug in decoding service types, when function types come after the service type
           in the type table
         </li>
+        <li>feat: support composite_query in candid</li>
         <li>
-          feat: support composite_query in candid fix: fix a bug in decoding service types, when
-          function types come after the service type in the type table
+          fix: fix a bug in decoding service types, when function types come after the service type
+          in the type table
         </li>
         <li>feat: include boundary node http details to query and update calls</li>
         <li>feat: adds method for actor creation that includes boundary node http details</li>
--- a/packages/agent/src/actor.ts
+++ b/packages/agent/src/actor.ts
@@ -2,6 +2,7 @@ import { Buffer } from 'buffer/';
 import {
   Agent,
   getDefaultAgent,
+  HttpDetailsResponse,
   QueryResponseRejected,
   QueryResponseStatus,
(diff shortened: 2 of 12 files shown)


docs/generated/changelog.html
```diff
@@ -32,9 +32,11 @@ <h2>Version x.x.x</h2>
           in the type table
         </li>
         <li>
-          feat: support composite_query in candid
-          fix: fix a bug in decoding service types, when function types come after the service type in the type table
+          feat: support composite_query in candid fix: fix a bug in decoding service types, when
```
dfx-json, line 35: Looks like the newline was accidentally removed here.

packages/agent/src/agent/api.ts
```diff
@@ -161,11 +172,16 @@ export interface Agent {
...
    * @param options Options to use to create and send the query.
+   * @param identity Sender principal to use when sending the query.
    * @returns The response from the replica. The Promise will only reject when the communication
    *     failed. If the query itself failed but no protocol errors happened, the response will
    *     be of type QueryResponseRejected.
    */
-  query(canisterId: Principal | string, options: QueryFields): Promise<QueryResponse>;
+  query(
+    canisterId: Principal | string,
+    options: QueryFields,
```
dfx-json, line 182: do you think it makes sense to add `identity` to `options` instead of changing the function signature?

packages/agent/src/actor.ts
```diff
@@ -285,11 +312,20 @@ export class Actor {
...
+    return new (this.createActorClass(interfaceFactory, false))(
       configuration,
     ) as unknown as ActorSubclass<T>;
   }
 
+  public static createActorWithHttpDetails<T = Record<string, ActorMethod>>(
+    interfaceFactory: IDL.InterfaceFactory,
+    configuration: ActorConfig,
+  ): ActorSubclass<ActorMethodMappedWithHttpDetails<T>> {
+    return new (this.createActorClass(interfaceFactory, true))(
```
rvanasa, line 324: It might be worth using an object parameter to simplify adding other configuration options in the future, i.e.:

```suggestion
    return new (this.createActorClass(interfaceFactory, {
      httpDetails: true,
    }))(
```

docs/generated/changelog.html
keplervital: This was done by the a pre-commit hook, might be related to having two types of changes under the same `li`, i'll one to another `li`

packages/agent/src/agent/api.ts
keplervital: I thought the same things when i was looking at this code, but i haven't modified anything, this was just added to the interface for completeness because the api query implementation was already receiving the identity as an extra argument [here](https://github.com/keplervital/agent-js/blob/8254c47694c1286058a8293547bda98a6d19c8e1/packages/agent/src/agent/http/index.ts#L377).

packages/agent/src/actor.ts
keplervital: Makes sense, updated it now 👍

dfx-json (approved):
LGTM. Thank you @keplervital !

PR #745, opened 2023-07-31
Feat/jsbi

Perlets9:
With react native there 2 big issues:

1. Bigint literals are not supported
2. WebAssembly is not supported

To address the first issue and support Bigint, is better to use a library, in this case, I chose the JSBI library.

Regarding the second issue, the web assembly is used by the bls verification. There is a possibility to use a custom version of the `blsVerify` as a parameter for the `pollForResponse` function, maybe using the @dfinity/bls-verify, but is not working at the moment. 

In this PR also there is a fix for the `pollForResponse` because, in some situations, wrong parameters were used.

### DIFF
--- a/demos/sample-javascript/package.json
+++ b/demos/sample-javascript/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@dfinity/ic-agent-sample-javascript-app",
   "private": true,
-  "version": "0.16.1",
+  "version": "0.18.1",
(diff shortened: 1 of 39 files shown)


sea-snake:
> 1. Bigint literals are not supported

BigInt are supported for a while already on React Native with `Hermes` engine, which has been the default since a few releases.

PR #747, opened 2023-08-01
feat: crypto nonce randomness

krpeacock:
# Description
 
Currently default nonce generation is using Math.random, a pseudorandom source of randomness. We can take advantage of better tools when they are available

Fixes SDK-1194

# How Has This Been Tested?

Updated and new unit tests

# Checklist:


packages/agent/src/agent/http/types.ts
```diff
@@ -109,8 +110,9 @@ export function makeNonce(): Nonce {
   const buffer = new ArrayBuffer(16);
   const view = new DataView(buffer);
   const now = BigInt(+Date.now());
-  const randHi = Math.floor(Math.random() * 0xffffffff);
-  const randLo = Math.floor(Math.random() * 0xffffffff);
+  const randHi = Math.floor(randomNumber() * 0xffffffff); //?
+  const randLo = Math.floor(randomNumber() * 0xffffffff); //?
```
ericswanson-dfinity, line 114: Suggestion: make it so we don't (in several cases) divide by 4294967295 only to multiply by 0xffffffff here
```suggestion
  const randHi = Math.floor(randomNumber()); //?
  const randLo = Math.floor(randomNumber()); //?
```

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,28 @@
+export const randomNumber = () => {
+  // determine whether browser crypto is available
+  if (typeof window !== 'undefined' && !!window.crypto && !!window.crypto.getRandomValues) {
+    const array = new Uint32Array(1);
+    window.crypto.getRandomValues(array);
+    return array[0] / 4294967295;
```
ericswanson-dfinity, line 6:
```suggestion
    return array[0];
```

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  // A second check for webcrypto, in case it is loaded under global instead of window
+  if (typeof crypto !== 'undefined' && crypto.getRandomValues) {
+    const array = new Uint32Array(1);
+    crypto.getRandomValues(array);
+    array;
+    array[0]; //?
+    return array[0] / 4_294_967_295;
```
ericswanson-dfinity, line 14:
```suggestion
    return array[0];
```

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  type nodeCrypto = {
+    randomInt: (min: number, max: number) => number;
+  };
+
+  // determine whether node crypto is available
+  if (typeof crypto !== 'undefined' && (crypto as unknown as nodeCrypto).randomInt) {
+    return (crypto as unknown as nodeCrypto).randomInt(0, 1_000_000_000) / 4_294_967_295;
```
ericswanson-dfinity, line 23:
```suggestion
    return (crypto as unknown as nodeCrypto).randomInt(0, 1_000_000_000);
```

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  // fall back to Math.random
+  return Math.random();
```
ericswanson-dfinity, line 27:
```suggestion
  return Math.random() * 0xffffffff;
```

packages/agent/src/utils/random.ts
ericswanson-dfinity, line 23: Also is 1_000_000_000 correct here?

krpeacock:
I was aiming initially to reproduce the behavior of Math.random, but I agree that yielding a random integer is preferable. I'll simplify, using your recommendations

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Generates a random unsigned 32-bit integer between 0 and 0xffffffff
+ * @returns {number} a random number
+ */
+export const randomNumber = () => {
```
ericswanson-dfinity, line 5: appropriate/useful?
```suggestion
export const randomNumber = (): number => {
```

packages/agent/src/utils/random.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
ericswanson-dfinity, line 1: Could you write some rationale here for why we are not requiring some well-established crypto library and always using it here?

```
import { randomBytes } from 'crypto';

export const randomNumber = (): number => {
  return randomBytes(4).readUInt32BE(0);
};
```

packages/agent/src/utils/random.ts
krpeacock: This is possibly the most controversial question in the TypeScript community 😁. I've chosen not to enforce return types for this codebase. 

Broadly speaking, this type is trivially inferred, so it does not add any utility for someone importing the function

packages/agent/src/utils/random.ts
ericswanson-dfinity: Thanks!

PR #766, opened 2023-09-20
feat: Principal class serializes to JSON

krpeacock:
# Description

`JSON.stringify(principal)` currently uses `Uint8Array` encoding instead of `toString`. Adding the `toJSON` method resolves this

# How Has This Been Tested?

new unit tests

# Checklist:

### DIFF
--- a/packages/principal/src/index.ts
+++ b/packages/principal/src/index.ts
@@ -108,6 +108,14 @@ export class Principal {
   }
 
   /**
+   * Serializes to JSON
+   * @returns {string} string
+   */
+  public toJSON(): string {
+    return this.toText();
+  }
+
+  /**
    * Utility method taking a Principal to compare against. Used for determining canister ranges in certificate verification
    * @param {Principal} other - a {@link Principal} to compare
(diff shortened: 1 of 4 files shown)


ericswanson-dfinity (approved):

packages/principal/src/index.ts
```diff
@@ -107,6 +107,14 @@ export class Principal {
     return this.toText();
   }
 
+  /**
+   * Serializes to JSON
+   * @returns {string} string
+   */
+  public toJSON(): string {
+    return this.toText();
```
peterpeterparker, line 115: Isn't this a bit weak? It means that any string is either a principal or a string which can be quite error prone if you stringify and parse an object that contains string and principals.

In `@dfinity/utils` we prefix principal to make their encoding and decoding more safe.

https://github.com/dfinity/ic-js/blob/main/packages/utils/src/utils/json.utils.ts

krpeacock: We could serialize into `{ "__principal__": principal.toText() }` instead. The main case I wanted to cover was just ensuring that `JSON.stringify()` on objects containing principals handles the case gracefully

PR #774, opened 2023-09-26
feat: retry logic catches thrown errors

krpeacock:
# Description

Currently, the request and retry feature only handles successful `fetch `requests that contain error responses. If the fetch request fails due to network conditions, the whole process ejects. This change will have the agent retry failed fetches

# How Has This Been Tested?

new Unit test

# Checklist:


krpeacock:
Next step will be adding an exponential backoff strategy to retries

packages/agent/src/agent/http/index.ts
```diff
@@ -361,28 +361,41 @@ export class HttpAgent implements Agent {
...
-      statusText: response.statusText,
-      headers: httpHeadersTransform(response.headers),
-    });
+      throw new AgentHTTPResponseError(errorMessage, {
+        ok: response.ok,
+        status: response.status,
+        statusText: response.statusText,
+        headers: httpHeadersTransform(response.headers),
+      });
+    } catch (error) {
```
ericswanson-dfinity, line 387: It might make sense to put this try..catch around only the `await request()`, or otherwise consolidate it with the `response.ok` checking.  Consider what happens now after throw new AgentHTTPResponseError().

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #776, opened 2023-09-28
feat: fetch node keys from subnet certificate

krpeacock:
# Description

Adds support for pulling subnet ID and node public keys out of a Certificate, and making them available for the `subnet` path in CanisterStatus.

Involves some refactoring of certificate decoding, so we introduce `lookupResultToBuffer` to constrain the `Cert.lookup` method to its existing return types. This means that a `HashTree` with type empty will be returned as `undefined`, preserving existing behavior. However, the `lookup_path` function will now return `HashTree`s.

This will support the upcoming certified queries feature

Fixes # (issue)

# How Has This Been Tested?

new tests, including fresh certificates from application and system subnets on local and mainnet, on `dfx 0.15.0`

# Checklist:


packages/agent/src/agent/http/__certificates__/goldenCertificates.test.ts
dfx-json, line 1: are these like the golden tickets in willy wonka?

packages/agent/src/certificate.ts
```diff
@@ -21,20 +21,35 @@ interface Cert {
...
+
+export { NodeId };
 
 export type HashTree =
-  | [NodeId.Empty]
-  | [NodeId.Fork, HashTree, HashTree]
-  | [NodeId.Labeled, ArrayBuffer, HashTree]
-  | [NodeId.Leaf, ArrayBuffer]
-  | [NodeId.Pruned, ArrayBuffer];
+  | [typeof NodeId.Empty]
```
dfx-json, line 37: how did this not produce a type error before?

packages/agent/src/certificate.ts
```diff
@@ -59,19 +74,38 @@ export function hashTreeToString(tree: HashTree): string {
...
+        return `sub(\n left:\n${indent(left)}\n---\n right:\n${indent(right)}\n)`;
+      } else {
+        throw new Error('Invalid tree structure for fork');
+      }
     }
     case NodeId.Labeled: {
-      const label = labelToString(tree[1]);
-      const sub = hashTreeToString(tree[2]);
-      return `label(\n label:\n${indent(label)}\n sub:\n${indent(sub)}\n)`;
+      if (tree[1] instanceof ArrayBuffer && tree[2] instanceof ArrayBuffer) {
```
dfx-json, line 86: some links to the spec as a comment would be helpful here imo

dfx-json (approved):
Looks good. Nice work @krpeacock 👍

packages/agent/src/certificate.ts
krpeacock: The JS agent's decoding logic simply didn't cover this case. The `/subnet` path never decoded correctly. 

At some point, the entire certificate ought to be re-written, modeled after the rust library for comprehensiveness

packages/agent/src/certificate.ts
krpeacock: This entire utility function doesn't seem to work and it has no tests. I only modified it in order to make it type safe with the other lookup changes

packages/agent/src/certificate.ts
```diff
@@ -172,7 +208,50 @@ export class Certificate {
...
+          'tdb26-jop6k-aogll-7ltgs-eruif-6kk7m-qpktf-gdiqx-mxtrf-vb5e6-eqe',
+        ).toUint8Array(),
+        certificate: new ArrayBuffer(0),
+      };
+    }
+    const nodeTree = lookup_path(['subnet', delegation?.subnet_id as ArrayBuffer, 'node'], tree);
+    const nodeForks = flatten_forks(nodeTree as HashTree) as HashTree[];
+    nodeForks.length;
+
+    this.#nodeKeys = nodeForks.map(fork => {
```
ShuoWangNSL, line 242: Here `nodeKeys` is an array of public keys. 
When we verify the node signature, we know the node id and would like to find the corresponding public. For this purpose, should `nodeKeys`  be a map from node id to public key?

PR #784, opened 2023-10-20
feat: node signature verification for queries

krpeacock:
# Description

In order to defend against malicious boundary nodes, subnet nodes will sign query responses. An Agent should follow the Internet Computer specification to validate these signatures

Fixes SDK-1216

# How Has This Been Tested?

# Checklist:


krpeacock:
oh thank god

packages/agent/src/agent/http/index.ts
```diff
@@ -407,63 +417,167 @@ export class HttpAgent implements Agent {
...
+   * See https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-query for details on validation
+   * @param queryResponse - The response from the query
+   * @param subnetStatus - The subnet status, including all node keys
+   * @returns ApiQueryResponse
+   */
+  #verifyQueryResponse = (
+    queryResponse: ApiQueryResponse,
+    subnetStatus: SubnetStatus | void,
+  ): ApiQueryResponse => {
+    if (!subnetStatus || this.#verifyQuerySignatures === false) {
```
ShuoWangNSL, line 527: It seems the only usage of `#verifyQueryResponse` is line 514. When would be case that `subnetStatus` is void?

packages/agent/src/agent/http/index.ts
```diff
@@ -407,63 +417,167 @@ export class HttpAgent implements Agent {
...
-      this._fetch('' + new URL(`/api/v2/canister/${canister.toText()}/query`, this._host), {
-        ...this._fetchOptions,
-        ...transformedRequest.request,
-        body,
-      }),
-    );
+      const separatorWithHash = concat(domainSeparator, new Uint8Array(hash));
 
-    const queryResponse: QueryResponse = cbor.decode(await response.arrayBuffer());
+      // FIX: check for match without verifying N times
```
ShuoWangNSL, line 563: Are we going to fix this in this PR?

packages/agent/src/agent/http/index.ts
```diff
@@ -407,63 +417,167 @@ export class HttpAgent implements Agent {
...
+      const matchingKey = subnetStatus?.nodeKeys.find(key => {
+        const pubKey = new Uint8Array(fromHex(key).slice(12, 44));
```
ShuoWangNSL, line 565: What about defining a method `from_der` somewhere?

packages/agent/src/canisterStatus/index.ts
```diff
@@ -110,10 +114,9 @@ export const request = async (options: {
           canisterId: canisterId,
         });
 
-        response.certificate;
         const lookup = (cert: Certificate, path: Path) => {
           if (path === 'subnet') {
-            const data = cert.cache_node_keys();
+            const data = fetchNodeKeys(response.certificate, agent.rootKey);
```
ShuoWangNSL, line 119: `response.certificate.verify()` is done inside the  `Certificate.create()`,right? In addition to the common verification for certificates, here we need extra check of canister ranges for certificate produced by NNS subnet. 
At this point, we already know the `path === 'subnet'` thus the path canister_ranges is directly included in the  certificate tree and we can perform similar check of canister ranges as we usually do in the delegation.

packages/agent/src/certificate.ts
```diff
@@ -216,44 +215,6 @@ export class Certificate {
...
-      }
-    });
-
-    return {
-      subnetId: Principal.fromUint8Array(new Uint8Array(delegation.subnet_id)).toText(),
-      nodeKeys: this.#nodeKeys,
-    };
-  }
-
   private async verify(): Promise<void> {
```
ShuoWangNSL, line 218: Are we going to disable the freshness check for delegation in this PR? If so, I guess it should be some changes in this function, right?

packages/agent/src/agent/http/index.ts
```diff
@@ -120,6 +121,11 @@ export interface HttpAgentOptions {
    * @default 3
    */
   retryTimes?: number;
+  /**
+   * Whether the agent should verify signatures signed by node keys on query responses. Increases security, but adds overhead and must make a separate request to cache the node keys for the canister's subnet.
+   * @default true
+   */
+  verifyQuerySignatures?: boolean;
```
ShuoWangNSL, line 128: It seems setting verifyQuerySignatures to false only skip verifying signature but not fetching node keys, right?

ShuoWangNSL: I cannot comment on line 118 about `disableNonce` so I put it here. 
`disableNonce` is still false by default. When and how we make it true by default for only queries?

packages/agent/src/agent/http/index.ts
ShuoWangNSL: I got some feedback from Franz-Stefan of the Crypto team. Maybe it is worth adding some more checks on the Der-encoding here.

> most likely, the basis of this decoding is a DER-encoded ASN.1 structure called SubjectPublicKeyInfo defined in [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280). This structure contains an algorithm identifier, and the public key itself in the subjectPublicKey field. If the public key is an Ed25519 public key, then according to [RFC 8410](https://datatracker.ietf.org/doc/html/rfc8410) the algorithm identifier must be 1.3.101.112.

> While the approach you are using of just extracting the key from the expected location generally works, it would be better if you also ensure that the algorithm ID is really correct. The DER-encoding also carries information of how long the contained subjectPublicKey is, which you could also check, which would give additional assurance that you are dealing with the correct/expected data.

> I don’t know the Javascript world so well, but ideally you would use some ASN.1 DER decoder and make sure all the parts of the DER-encoding have the expected format.

packages/agent/src/agent/http/index.ts
krpeacock: This is resolved in #792

packages/agent/src/canisterStatus/index.ts
krpeacock: I'm deliberately moving some of this logic out of the certificate, because I don't want to overload the responsibility of the certificate class with more specific validations

krpeacock: The certificate is already validated on line 119. FetchNodeKeys is just responsible for parsing the rest of the subnet metadata out of the state tree

ShuoWangNSL: > The certificate is already validated on line 119, and will verify the delegation, including the canister ranges.

That's right for certificate with a delegation. I am talking about the special case of a certificate produced by the root subnet. It has no delegation and thus the canister ranges are not verified. Therefore, in addition to the common verification done to certificates, an extra check of canister ranges needs to be done when we fetch node keys in root subnet.

ShuoWangNSL: We discussed it some long time ago. Here are the relevant design [slides](https://docs.google.com/presentation/d/1xbobmOhplo-OwjP41JY5MhpI7c-peyJPtNE-v_-DJKo/edit?disco=AAAA-gP82Ik).

packages/agent/src/agent/http/index.ts
krpeacock: I'd suggest a re-read. VerifyQueryResponse is a new configuration option that defaults to true. SubnetStatus can be void if the developer deliberately opts out of verification, similar to what agent-rs is doing

packages/agent/src/agent/http/index.ts
krpeacock: Can you explain further? I don't recall that part of the spec

packages/agent/src/canisterStatus/index.ts
ShuoWangNSL: Here is the example.
Suppose the agent-js is making a query call to canister C on an App subnet A for the first time. Because it has no cached node keys, it sends out a read_state call requesting `/subnet` simultaneously.  However, the malicious boundary node route the read_state call and query call to a malicious node N in the root subnet. Node N returns a real certificate and a forged query response.

For the agent's perspective, the certificate is correct and it will use the node N's public key to verify the query response. The problem is that without checking the canister ranges in the certificate, it does not know that node N on the root subnet is not authorized to answer calls for canister C. That's why we request the whole `/subnet` path in addition to the node key path. In this way we can verify canister ranges directly from the tree of certificate when the delegation is absent for the root subnet.

I understand that this extra check is a special requirement and may fit in the current code perfectly. However we need it because one of feature goals is  **"subnet Y is authoritative for the canister targeted with the query"**.

packages/agent/src/agent/http/index.ts
ShuoWangNSL: It seems `subnetStatusPromise` is always executed thus setting `verifyQuerySignatures` as false only skips verification but it still fetches node keys. The agent-rs code skips both.

packages/agent/src/agent/http/index.ts
ShuoWangNSL: Here we try all the node keys stored in `subnetStatus` to see if one of them can match the signature. 
The comment you left  `FIX: check for match without verifying N times` is from the previous PR about fetching node keys.
In the previous PR , we flatten the forks under `/subnet/<subnet_id>/node` and get a list of public keys. Because the path `/subnet/<subnet_id>/node/<node_id>/public_key` also contains the information of node id, we can store the mapping of `node-id->public key` in the `subnetStatus`. Then, instead of trying all keys, we just get it the from mapping by the `sig.identity`(the node_id which generated the response).

packages/agent/src/agent/http/index.ts
roelstorms: Just to clarify, #792 only solves the second comment but not the first comment, right?

> It seems setting verifyQuerySignatures to false only skip verifying signature but not fetching node keys, right?

roelstorms: Nvm, seems that fetching nodes keys is only done when verifyQuerySignatures is false. This is checked on line 511.

packages/agent/src/agent/http/index.ts
roelstorms: This is the same comment as given above.

IIUC, fetchSubnetKeys on line 518 is only executed if the check against #verifyQuerySignatures on line 511 fails.

That seems to ensure node keys are only fetched if #verifyQuerySignature is true.

roelstorms: Also on the subnetStatus being void, I agree that #verifyQueryResponse shouldn't be called with void the calling code seems to assure that. It's still good to check that it is actually the case inside #verifyQueryResponse.

fetchSubnetKeys seems to ensure that subnetStatus is not void.

However, if for some reason #verifyQueryResponse was called with this.#verifyQuerySignatures === true but subnetStatus void, it would not do any verification and just return the queryResponse.

Shouldn't verification fail in such a situation?

packages/agent/src/agent/http/index.ts
```diff
@@ -424,64 +435,172 @@ export class HttpAgent implements Agent {
...
 
-    const request: QueryRequest = {
-      request_type: ReadRequestType.Query,
-      canister_id: canister,
-      method_name: fields.methodName,
-      arg: fields.arg,
-      sender,
-      ingress_expiry: new Expiry(DEFAULT_INGRESS_EXPIRY_DELTA_IN_MSECS),
-    };
+      const canister = typeof canisterId === 'string' ? Principal.fromText(canisterId) : canisterId;
```
roelstorms, line 446: It looks like Principal.from can create a principal from string or principal.
Could this line be replaced with `Principal.from(canisterId);` to simplify it?

packages/agent/src/certificate.ts
roelstorms: verify is called as part of Certificate.create which is called on the certificate in the delegation.
verify checks that /time is past now - 5 min and before now + 5 min.

The IC returns root certificates which are several days or weeks old. This validation would fail. See [FOLLOW-1294 ](https://dfinity.atlassian.net/browse/FOLLOW-1294 )

packages/agent/src/agent/http/index.ts
krpeacock: It can
```suggestion
      const canister = Principal.from(canisterId);
```

packages/agent/src/agent/http/index.ts
krpeacock: I've accomplished this by bringing in our DER-encoding logic for ED25519 public keys from the @dfinity/identity package

packages/agent/src/agent/http/index.ts
```diff
@@ -424,63 +435,171 @@ export class HttpAgent implements Agent {
...
+      const { timestamp, identity } = sig;
+      const nodeId = Principal.fromUint8Array(identity).toText();
+      let hash: ArrayBuffer;
+
+      // Hash is constructed differently depending on the status
+      if (status === 'replied') {
+        const { reply } = queryResponse;
+        hash = hashOfMap({
+          status: status,
+          reply: hashOfMap(reply),
```
ShuoWangNSL, line 562: According to the latest specification and the final clarification on the `hash_of_map`, for nested map, we don't hash them twice. So instead of using `hashOfMap(reply)`, we just put `reply` here and adapt the `hashValue` function to support Record<string , unknown>  type and call `hash_of_map` there.

packages/agent/src/agent/http/index.ts
krpeacock: This is fixed in https://github.com/dfinity/agent-js/pull/784/commits/fba09bf956943688625bcc02c066848e63ce3478

packages/agent/src/agent/http/index.ts
krpeacock: Resolved. Hope this is the last adjustment!

packages/agent/src/agent/http/index.ts
krpeacock: Yes. In fact, fixing this check fixed a number of false positives in unit tests that ought to have verification disabled

packages/agent/src/certificate.ts
krpeacock: The code is a little confusing, but inside of a standard `certificate`, there are two `certificates`. The main one, and the one contained inside the `delegation`. on line 263, we set a special maxAgeInMinutes of 30 days, which hasn't seemed to fail against the mainnet in practice. We could extend this as far as we want though

https://github.com/dfinity/agent-js/blob/a899a3cadc92430071dce6dacfac1333be255e38/packages/agent/src/certificate.ts#L263

roelstorms: @ShuoWangNSL could you check whether the behavior is/should be aligned between js and rs.
Atm, I think js checks the cert against 5min **window** unless it's a cert inside a delegation where it's 30 days.
rs only checks the cert is not older than 5 min and does not check the time on certs found in delegations.

Eventually, all certs should use the same 5 min windows (see [FOLLOW-692](https://dfinity.atlassian.net/browse/FOLLOW-692))

krpeacock: As we discussed this morning, I agreed to remove the check. In code, this is done by setting the maxAge to `Infinity`

packages/agent/src/canisterStatus/index.ts
```diff
@@ -209,6 +234,69 @@ export const request = async (options: {
...
+    throw new Error('Canister not in range');
+  }
+
+  const nodeTree = lookup_path(['subnet', delegation?.subnet_id as ArrayBuffer, 'node'], tree);
+  const nodeForks = flatten_forks(nodeTree as HashTree) as HashTree[];
+  nodeForks.length;
+  const nodeKeys = new Map<string, DerEncodedPublicKey>();
+  nodeForks.forEach(fork => {
+    Object.getPrototypeOf(new Uint8Array(fork[1] as ArrayBuffer));
+    const node_id = Principal.from(new Uint8Array(fork[1] as ArrayBuffer)).toText();
```
ShuoWangNSL, line 285: We got error:
`index.js:403 Expected to find result for path subnet, but instead found nothing.
(anonymous) @ index.js:403
index.js:403 Error: Impossible to convert {"0":2,"1":229,"2":223,"3":21,"4":96,"5":194,"6":199,"7":168,"8":58,"9":105,"10":151,"11":120,"12":17,"13":148,"14":220,"15":20,"16":146,"17":229,"18":85,"19":55,"20":122,"21":97,"22":23,"23":146,"24":236,"25":191,"26":229,"27":1,"28":2} to Principal.`
Should here be`Principal.fromUint8Array()`? I am curious why it is not detected by tests.

krpeacock: I shipped a change in this PR that updates the `Principal.from` method as well, so that it will cover `Uint8Arrays`. You'll need to pack and install `@dfinity/principal`

dfx-json (approved):

PR #788, opened 2023-10-26
feat: round ingress expiry

krpeacock:
# Description

In order to more effectively provide caching options in the boundary nodes, it is useful to limit the areas of randomness among cacheable query results. This feature will set expiries to 5 minutes, rounded down to the nearest minute.

Fixes # (issue)

# How Has This Been Tested?

# Checklist:


krpeacock:
Note: This change makes it easy to reproduce what would have been an edge case before. By rounding the expiry, it means that sequential `update` calls may return the same response if the agent's `disableNonce` option is set.

This is not yet the default, so the scope of this PR doesn't need to cover that

dfx-json (approved):

PR #796, opened 2023-11-08
feat: uses expirable map for subnet keys

krpeacock:
# Description

Adds an expirable map so that node keys will be checked. 

Currently set to expire after an hour, but we could make it shorter

Fixes # (issue)

# How Has This Been Tested?

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -186,7 +187,9 @@ export class HttpAgent implements Agent {
   #queryPipeline: HttpAgentRequestTransformFn[] = [];
   #updatePipeline: HttpAgentRequestTransformFn[] = [];
 
-  #subnetKeys: Map<string, SubnetStatus> = new Map();
+  #subnetKeys: ExpirableMap<string, SubnetStatus> = new ExpirableMap({
```
ShuoWangNSL, line 190: Last time we [discussed](https://dfinity.slack.com/archives/CGA566TPV/p1699308907818999?thread_ts=1699306329.116199&cid=CGA566TPV) that the Map is ideally be
```
ExpirableMap<
  subnet_id,
  Map<canister_id, node_key>
>
```
so that different canisters in a subnet can share the cache.
However, given that query calls to different canisters are likely to use different instances of JS agent, the optimization seems unnecessary now. I agree we can keep this for now and consider using subnet-id as key when we use the IndexedDB in the future.

krpeacock: I thought the purpose of this is to ensure that the node keys are eventually re-checked, since the actual delegations will change in upgrades

ShuoWangNSL: Yes. The purpose of `ExpirableMap` is to ensure node keys are eventually re-checked.  It looks good to me.

What I mentioned above is about reusing subnet cache for different canisters. It can be regarded as out of the scope of this PR.

krpeacock: cool, this should be fine then. Putting the cache at the top level is simpler, but it will have the same desired effect as the node cache - it's just that the #subnetKeys instance is meant to live longer

dfx-json (approved):

PR #801, opened 2023-11-16
feat: retry query signature verification in case cache is stale

krpeacock:
# Description

Because a node can be upgraded while the cache is stale, we should retry a failed query signature verification once, after refreshing the cache

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -532,12 +532,9 @@ export class HttpAgent implements Agent {
     if (!this.#verifyQuerySignatures) {
       return query;
     }
-    console.log("here");
-    const isValid = this.#verifyQueryResponse(query, subnetStatus);
-    console.log("isValid 1", isValid);
-    if (isValid === true) {
-      return query;
-    } else {
+    try {
+      return this.#verifyQueryResponse(query, subnetStatus);
+    } catch (_) {
       // In case the node signatures have changed, refresh the subnet keys and try again
       console.warn('Query response verification failed. Retrying with fresh subnet keys.');
       this.#subnetKeys.delete(canisterId.toString());
@@ -549,13 +546,7 @@ export class HttpAgent implements Agent {
           'Invalid signature from replica signed query: no matching node key found.',
         );
       }
-      const isValid = this.#verifyQueryResponse(query, updatedSubnetStatus);
-      console.log('isValid 2', isValid);
-      if (isValid === true) {
-        return query;
-      } else {
-        throw isValid;
-      }
+      return this.#verifyQueryResponse(query, updatedSubnetStatus);
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/agent/src/agent/http/index.ts
```diff
@@ -532,7 +532,20 @@ export class HttpAgent implements Agent {
...
     }
-    return this.#verifyQueryResponse(query, subnetStatus);
+
+    try {
+      return this.#verifyQueryResponse(query, subnetStatus);
+    } catch (error) {
+      // In case the node signatures have changed, refresh the subnet keys and try again
+      console.warn('Query response verification failed. Retrying with fresh subnet keys.');
+      const nodeId = Principal.from(query.signatures?.[0].identity);
+      this.#subnetKeys.delete(nodeId.toText());
```
ShuoWangNSL, line 542: `subnetKeys` is a mapping from canister_id to `SubnetStatus`.

packages/agent/src/agent/http/index.ts
```diff
@@ -532,7 +532,20 @@ export class HttpAgent implements Agent {
...
+      await this.fetchSubnetKeys(nodeId);
```
ShuoWangNSL, line 543: Ditto.  `canisterId`.

packages/agent/src/agent/http/index.ts
```diff
@@ -532,7 +532,20 @@ export class HttpAgent implements Agent {
...
+
+      const updatedSubnetStatus = this.#subnetKeys.get(nodeId.toText());
```
ShuoWangNSL, line 545: Ditto.  `canisterId.toText()`.

krpeacock:
It's throwing an error, but not in a way that the mitm test is detecting as a "rejection" I need to investigate a bit further

packages/agent/src/agent/http/index.ts
```diff
@@ -532,7 +532,22 @@ export class HttpAgent implements Agent {
...
+      this.#subnetKeys.delete(canisterId.toString());
+      await this.fetchSubnetKeys(canisterId.toString());
+
+      const updatedSubnetStatus = this.#subnetKeys.get(canisterId.toString());
+      if (!updatedSubnetStatus) {
+        throw new CertificateVerificationError(
+          'Invalid signature from replica signed query: no matching node key found.',
+        );
+      }
+      return this.#verifyQueryResponse(query, updatedSubnetStatus);
```
ericswanson-dfinity, line 549: Is it okay that there is no `resolve(subnetStatus);` call here, in between `get` and `verifyQueryResponse`, as in the above code?

krpeacock: Yes, we are not within a `Promise` at this stage in the code (apart from the `async query` function itself). It's a little semantically incongruous if you're comparing this section to the above sections, however

PR #816, opened 2024-01-03
feat: adds fromPem method for identity-secp256k1

krpeacock:
# Description

Adds a utility method to initialize a secp256k1 identity from a pem file.

Fixes SDK-1255
Fixes SDKTG-221

# How Has This Been Tested?

Tested against a test identity exported from `dfx` and compatible with `dfx` and `keysmith`

# Checklist:


packages/identity-secp256k1/src/pem.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (rawKey.length !== 118) {
+      throw new Error(`Invalid key length ${rawKey.length}. Expected 118 bytes.`);
+    } else {
+      return rawKey.slice(7, 39);
+    }
+  }
+  if (rawKey.length != 85) {
+    throw new Error(`Invalid key length ${rawKey.length}. Expected 85 bytes.`);
+  } else {
+    return concat(rawKey.slice(16, 48), rawKey.slice(53, 85));
```
randombit, line 38: I have no idea what is happening here :)

It seems like this is assuming that if the PEM header is not "EC PRIVATE KEY" then it is some other specific unnamed format, and that performing this operation where we remove (?) 5 bytes from the middle is appropriate.

At the very least this deserves a comment regarding what data is being removed, and probably a check that the header is the expected one.

packages/identity-secp256k1/src/pem.ts
```diff
@@ -0,0 +1,42 @@
...
+  const footer = lines[lines.length - 1].trim();
+  if (lines.length < 3) {
+    throw new Error('Invalid PEM format');
+  }
+  if (!header.startsWith(HEADER)) {
+    throw new Error('Invalid PEM header');
+  }
+  if (!footer.startsWith(FOOTER)) {
+    throw new Error('Invalid PEM footer');
+  }
```
randombit, line 24: This will accept something like

```
----BEGIN EC PRIVATE KEY-----
// appropriate base64 encoding
-----END SURPRISE-----
```

or really even

```
----BEGIN EC PRIVATE KEY-----
// appropriate base64 encoding
-----END 
```

Generally you see functions like this operate in one of two ways:

* Takes the input string, plus the expected PEM label (eg "EC PRIVATE KEY"), and return the decoded bits. If the label doesn't match (checking both header and footer), return an error. This is useful for cases where you know what the PEM label is supposed to be.
* Take the input string, and return both the decoded bits plus the sniffed label (verifying it is used for both header + footer). This is useful where you might have to accept arbitrary labels, say for a pretty printer.

Often if the first style is used, it is defined in terms of the second style, eg (apologies for my psuedo-JS):

```
// TS has tuples right???
const (bits, label) = pem_decode(pemStr);
if label != expectedLabel {
  throw new Error(...);
}
return bits;
```

IIUC you only need the first style here.

packages/identity-secp256k1/src/pem.ts
```diff
@@ -0,0 +1,42 @@
...
+  const base64Data = lines.slice(1, -1).join('').replace(/\r?\n/g, '');
+  const rawKey = Uint8Array.from(atob(base64Data), c => c.charCodeAt(0));
+
+  if (pem.startsWith(ECHEADER)) {
+    if (rawKey.length !== 118) {
```
randombit, line 29: This seems somewhat a tangling of concerns. I would expect a length check to happen in code that decodes the binary data (here, fromSecretKey)

packages/identity-secp256k1/src/pem.ts
```diff
@@ -0,0 +1,42 @@
...
+      throw new Error(`Invalid key length ${rawKey.length}. Expected 118 bytes.`);
+    } else {
+      return rawKey.slice(7, 39);
```
randombit, line 32: I guess this is just extracting the raw 32 byte binary key at the appropriate offset? Probably worth a comment.

packages/identity-secp256k1/src/pem.ts
randombit: Oh I see from the tests I'm guessing this is `PRIVATE KEY`

A comment on the format and which bits are being extracted seems warranted.

Also should point out that `PRIVATE KEY` is in practice used for many different kinds of key, so if you put an RSA private key, or even just an encrypted ECDSA key, this function will just return some junk. If supporting this format is necessary you should probably at a minimum check the leading bytes, which will contain a key identifier

```
    0:d=0  hl=3 l= 132 cons: SEQUENCE          
    3:d=1  hl=2 l=   1 prim: INTEGER           :00
    6:d=1  hl=2 l=  16 cons: SEQUENCE          
    8:d=2  hl=2 l=   7 prim: OBJECT            :id-ecPublicKey
   17:d=2  hl=2 l=   5 prim: OBJECT            :secp256k1
   24:d=1  hl=2 l= 109 prim: OCTET STRING      [HEX DUMP]:....
```

krpeacock:
@randombit got back to this today - should be more legible now. Thanks for your suggestions!

docs/generated/changelog.html
```diff
@@ -40,7 +41,9 @@ <h2>Version 0.21.0</h2>
         <li>feat: introduces Observable Log to HttpAgent. Allows subscribers to be notified of events from the agent without sending them directly to the console</li>
         <li>chore: exporting http errors</li>
         <li>chore: cleaning up lint warnings</li>
+>>>>>>> main
```
dfx-json, line 44: leftover from merge?

dfx-json (approved):

docs/generated/changelog.html
krpeacock, line 44:
```suggestion
```

docs/generated/changelog.html
```diff
@@ -40,7 +41,9 @@ <h2>Version 0.21.0</h2>
...
         <li>chore: cleans up github actions linting warnings</li>
+>>>>>>> main
```
krpeacock, line 45:
```suggestion
```

randombit:
LGTM

PR #817, opened 2024-01-08
feat: pure JS BLS verification

krpeacock:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

# How Has This Been Tested?

# Checklist:


krpeacock:
After investigating, I confirmed that [this commit](https://github.com/paulmillr/noble-curves/commit/2f1460a4d7f31e5a61db9606266f5b9a5c659c9d) broke the noble curves short signature implementation shortly before the release. If we fix the issue, this PR should be ready to ship once the next version of `@noble/curves` containing the patch is out

krpeacock:
> I don't see any _new_ tests added. Are we reasonably confident the noble/curves implementation is secure and provides parity with the existing?

@randombit added support for the library, and also authored the latest patch that restored functionality -  https://github.com/paulmillr/noble-curves/issues/124

The existing tests, including tests against mainnet, should be adequate. Happy to pass this to prodsec for additional review.

Additionally, we could pin the version for key crypto dependencies like this, and require manual bumping for any changes, including patches

packages/agent/src/certificate.ts
```diff
@@ -122,7 +122,7 @@ function isBufferEqual(a: ArrayBuffer, b: ArrayBuffer): boolean {
   return true;
 }
 
-type VerifyFunc = (pk: Uint8Array, sig: Uint8Array, msg: Uint8Array) => Promise<boolean>;
+type VerifyFunc = (pk: Uint8Array, sig: Uint8Array, msg: Uint8Array) => Promise<boolean> | boolean;
```
ericswanson-dfinity, line 125: Is this  (`=> Promise<boolean> | boolean`) for backwards compatibility?  An approach I've seen elsewhere is to return a completed future in cases like this.

krpeacock: It's to allow an interface to have the option to make a call synchronously. The new library is synchronous, and the Promise is the backwards compatibility

dfx-json (approved):

PR #854, opened 2024-02-21
feat: replay attack prevention using watermarks

krpeacock:
# Description

Several [reports](https://taggr.link/#/thread/43171) have been received about an issue affecting how immediately a response from an update call can be used. One example is the following scenario:

an update call is made to create a new resource (e.g. a social media post)

the application navigates to a page that displays the resource (e.g. the post itself)

the application requests the resource by executing a query call and using the resource id 

the query hits a replica whos state has not yet been updated

a 404 is returned

the problem is, the resource does exist, it just so happens that 1/3 of replicas may not necessarily be up to date at the same moment the other 2/3’s are

Fixes # SDK-1332

# How Has This Been Tested?

New unit and end to end tests, introducing a replay attack test in `e2e/node/basic/watermark.test.ts`

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -418,13 +425,98 @@ export class HttpAgent implements Agent {
...
+    // Convert the timestamp to milliseconds
+    const timeStampInMs = Number(BigInt(timestamp) / BigInt(1_000_000));
+
+    this.log('watermark and timestamp', {
+      waterMark: this.waterMark,
+      timestamp: timeStampInMs,
+    });
+
+    // If the timestamp is less than the watermark, retry the request up to the retry limit
+    if (timestamp && Number(this.waterMark) > timeStampInMs) {
```
chmllr, line 494: nit: do we even need to evaluate `timestamp` at that point? We checked it in line 479, right?

chmllr (approved):
Thanks so much Kyle! Please announce the release with this fix on Taggr, I'll give a ⭐️! 😅

packages/auth-client/test-setup.ts
```diff
@@ -18,11 +18,7 @@ Object.defineProperty(globalThis, 'crypto', {
   value: crypto,
 });
 
-Object.defineProperty(console, 'warn', {
-  value: jest.fn(),
+Object.defineProperty(global, 'console', {
+  writable: true,
+  value: { ...global.console, log: jest.fn(), warn: jest.fn(), error: jest.fn() },
```
dfx-json, line 23: does this mean that no logs will be printed during test execution?

packages/agent/src/agent/http/index.ts
krpeacock: Yes - the error above makes this redundant. I'll simplify

packages/auth-client/test-setup.ts
krpeacock: Yes, it's to clean up the console by default when running the tests. However, the logs aren't lost and you can access them as needed

dfx-json (approved):

PR #856, opened 2024-03-14
feat: support for restricting II auth methods

krpeacock:
# Description

Adds new security features from the II login flow for restricting authentication types: https://github.com/dfinity/internet-identity/blob/6c345912457ca93bf3175ce3382a1ddf8ff46eca/docs/ii-spec.md#client-authentication-protocol

New Login method allows developers to restrict users from authenticating using a PIN
image

Also allows `onSuccess` to pass through the message from the IDP provider, allowing developers to access the new `authnMethod` field

image

Fixes # (issue)

# How Has This Been Tested?

Tested manually after building and installing in https://github.com/krpeacock/auth-client-demo

# Checklist:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,7 @@
     * New login option: `allowPinAuthentication?: boolean;`
     * Response from II includes `authnMethod: 'passkey' | 'pin' | 'recovery';`
     * OnSuccess now optionally passes the message directly from the IDP provider
+    * Support for arbitrary login values passed to IDP through `customValues` option
 
 ### Added
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Changed
+
+* feat!: support for restricting II auth methods
+    * New login option: `allowPinAuthentication?: boolean;`
+    * Response from II includes `authnMethod: 'passkey' | 'pin' | 'recovery';`
+    * OnSuccess now optionally passes the message directly from the IDP provider
+
 ### Added
 
 * feat: adds `fromPem` method for `identity-secp256k1`
--- a/packages/auth-client/src/index.ts
+++ b/packages/auth-client/src/index.ts
@@ -97,6 +97,10 @@ export interface AuthClientLoginOptions {
    */
   maxTimeToLive?: bigint;
   /**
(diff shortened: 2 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Changed
+
+* feat!: support for restricting II auth methods
+    * New login option: `allowPinAuthentication?: boolean;`
+    * Response from II includes `authnMethod: 'passkey' | 'pin' | 'recovery';`
+    * OnSuccess now optionally passes the message directly from the IDP provider
```
nmattia, line 10: very nice! just a thought: also allowing extra args in `login()` (sent directly to the IDP provider) would also be nice

packages/auth-client/src/index.ts
```diff
@@ -463,7 +475,11 @@ export class AuthClient {
     /**
      * Callback once login has completed
      */
-    onSuccess?: (() => void) | (() => Promise<void>);
+    onSuccess?:
+      | (() => void)
+      | ((message?: InternetIdentityAuthResponseSuccess) => void)
+      | (() => Promise<void>)
+      | ((message?: InternetIdentityAuthResponseSuccess) => Promise<void>);
```
nmattia, line 482: making the `message` optional means the user cannot actually rely on `message` existing I think! how about something like this?

`(message: IIAuthResponse) => void | Promise<void>`

that should represent the same type, though will prevent the user from performing an unnecessary undefined check

krpeacock:
```suggestion
      | (() => void | Promise<void>)
      | (message: IIAuthResponse) => void | Promise<void>);
```
I think that we can do that, and also offer the same version with no message, like this. It'll be half as verbose

krpeacock: I had a test that failed because the signature changed, which signaled to me that supporting the zero argument signature is important for compatibility

docs/CHANGELOG.md
krpeacock: As in arbitrary args? I could add that

packages/auth-client/src/index.ts
nmattia: Ok, quite surprising! Won't this force the users to check for undefined even though the argument will never be undefined?

nmattia (approved):
Looking great, thanks! Have you tried it out? I think the feature's live in II already

krpeacock:
> Looking great, thanks! Have you tried it out? I think the feature's live in II already

Yep - the screenshot was from mainnet

ericswanson-dfinity (approved):

PR #871, opened 2024-04-02
feat: retry delay strategy

krpeacock:
# Description

Developers have noticed the agent is more frequently erroring with the new watermark protections against replay attacks / stale data. This feature adds a delay strategy for retries that will allow for more time for nodes to catch up, with exponential increases to the rate

Fixes SDK-1562

# How Has This Been Tested?

new e2e tests

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -124,6 +125,10 @@ export interface HttpAgentOptions {
    * @default 3
    */
   retryTimes?: number;
+  /**
+   * The fetch implementation to use. Defaults to the global fetch function.
+   */
+  delayStrategy?: BackoffStrategy;
```
ericswanson-dfinity, line 131: These names could be consistent, either delayStrategy / DelayStrategy or backoffStrategy / BackoffStrategy
```suggestion
  backoffStrategy?: BackoffStrategy;
```

Also the comment ("fetch implementation") looks like it was copy/pasted from elsewhere.

packages/agent/src/polling/strategy.ts
```diff
@@ -135,3 +135,33 @@ export function chain(...strategies: PollStrategy[]): PollStrategy {
...
+
+/**
+ * An exponential backoff strategy.
+ * @param iterations The number of retries that have been done.
+ */
+export function exponentialBackoff(iterations: number): Delay {
+  if (iterations === 0) {
+    return 0;
+  }
+  return 2 ** iterations * 150;
```
ericswanson-dfinity, line 151: I recommend introducing jitter here.

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Exponential backoff policy that we use in dfx and the rust agent: https://github.com/ihrwein/backoff/blob/master/src/exponential.rs#L111

ericswanson-dfinity (changes requested):

packages/agent/src/polling/strategy.ts
krpeacock: Jitter is interesting - it'll make the tests a little more complicated, but I see the utility of it

ByronBecker (commented):
Exponential backoff with jitter should work fine for most cases, but assuming no jitter in the case of 3 retries that's a call at 0, 150ms, 600ms, and 1200ms (am I correct here?)

It would be nice to see if this covers every case, or if there would still be a super small chance that this error might get hit (even with the exponential backoff).

> Timestamp failed to pass the watermark after retrying the configured 3 times. We cannot guarantee the integrity of the response since it could be a replay attack.

If this error got hit with the exponential backoff in place, what would that mean for the FE app? Is the subnet down or behind? Did we just get super unlucky? Does the same error message make sense now that the exponential backoff is in place (i.e. does this error have a different meaning)?

Ideally, this error would never happen by accident, or even if the FE client user has a poor internet connection. Then if they receive the error it would actually suggest something is not healthy with the nodes/subnet they've communicated with.

Putting a higher threshold in by using exponential backoff and hoping that this is a rare case is just going to confuse developers, and especially confuse end users in the even this error is hit and a toast component pops up with "Timestamp failed to pass the watermark after retrying the configured 3 times. We cannot guarantee the integrity of the response since it could be a replay attack" (in fact, this sort of sounds like security issue).

packages/agent/src/agent/http/index.ts
```diff
@@ -269,7 +274,13 @@ export class HttpAgent implements Agent {
...
     }
     // Default is 3
-    this._retryTimes = options.retryTimes ?? 3;
+    this.#retryTimes = options.retryTimes ?? 3;
+    // Delay strategy for retries. Default is exponential backoff
+    const defaultBackoffFactory = () =>
+      new ExponentialBackoff({
+        maxIterations: this.#retryTimes,
+      });
+    this.#backoffStrategy = options.backoffStrategy ?? defaultBackoffFactory;
```
dfx-json, line 283: is the typescript compiler enough to ensure `options.backoffStrategy` isnt something other than a function?

dfx-json (approved):
LGTM

packages/agent/src/agent/http/index.ts
krpeacock: No, it wouldn't be prevented, but we aren't doing strict type enforcement on most options during construction in this code base at this stage

dfx-json: in this case would `||` be a better fit in case someone provides e.g. `options.backoffStrategy = false` ?

krpeacock: fair enough
```suggestion
    this.#backoffStrategy = options.backoffStrategy || defaultBackoffFactory;
```

PR #873, opened 2024-04-12
feat: deprecate `HttpAgent` constructor in favor of new `create`

krpeacock:
# Description

Following the pattern used by other interfaces in the library, this function switches to an async `create` method, that will be able to make the `syncTime` and `fetchRootKey` calls before completing. This will also support future async setup calls if needed in the future. `createSync` initializes the agent but skips any async steps, and is useful for testing or more sophisticated applications

The agent also has a `from` method, and will support either the v1 or v2 agent interfaces. 

The HttpAgent also switches to true es6 private class fields, and will formally reveal its config and its host as fields. This supports the from method, but the `_`-prefixed methods from older HttpAgents will be supported.

> Note: the constructor is marked deprecated in this PR. Existing code will not break yet, but will inform the developer that they should switch to `create or `createSync`.

# How Has This Been Tested?

# Checklist:


krpeacock:
@roelstorms

dfx-json (approved):

PR #874, opened 2024-04-25
chore: Add context to errors thrown on cbor decode

frederikrothenberger:
We have the issue in Internet Identity that people complain about hard to diagnose issues. In particular, we regularly encounter users faced with the "failed to parse" message thrown out of `borc`. It would be very helpful to know what the input was that caused that error.

# How Has This Been Tested?

Against existing test suite.

# Checklist:

### DIFF
--- a/packages/agent/src/cbor.ts
+++ b/packages/agent/src/cbor.ts
@@ -4,7 +4,7 @@ import borc from 'borc';
 import * as cbor from 'simple-cbor';
 import { CborEncoder, SelfDescribeCborSerializer } from 'simple-cbor';
 import { Principal } from '@dfinity/principal';
-import { concat, fromHex } from './utils/buffer';
+import { concat, fromHex, toHex } from './utils/buffer';
 
 // We are using hansl/simple-cbor for CBOR serialization, to avoid issues with
 // encoding the uint64 values that the HTTP handler of the client expects for
@@ -125,5 +125,9 @@ export function decode<T>(input: ArrayBuffer): T {
     },
   });
 
-  return decoder.decodeFirst(buffer);
+  try {
+    return decoder.decodeFirst(buffer);
+  } catch(e: unknown) {
+    throw new Error(`Failed to decode CBOR: ${e}, input: ${toHex(buffer)}`);
+  }
 }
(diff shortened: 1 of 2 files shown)


krpeacock (approved):

packages/agent/src/cbor.ts
```diff
@@ -125,5 +125,9 @@ export function decode<T>(input: ArrayBuffer): T {
     },
   });
 
-  return decoder.decodeFirst(buffer);
+  try {
+    return decoder.decodeFirst(buffer);
+  } catch(e: unknown) {
+    throw new Error(`Failed to decode CBOR: ${e}, input: ${toHex(buffer)}`);
```
peterpeterparker, line 131: Out of curiosity, aren't you gonna get something like `Failed to decode CBOR: [object Object]...` - i.e. `e` is always a string?

frederikrothenberger: Oh, thanks. I think having the `input: ...` part is definitely an improvement though.
So, what's the simplest way of converting `unknown` to something readable?

peterpeterparker: Not sure what's the best practice or if there is an elegant solution.

In Oisy I got such a parser:

```
export const errorDetailToString = (err: unknown): string | undefined =>
	typeof err === 'string'
		? (err as string)
		: err instanceof Error
			? (err as Error).message
			: 'message' in (err as unknown as { message: string })
				? (err as unknown as { message: string }).message
				: undefined;
```

Another option is maybe to create an error type that extends the error and bubble the new types with the information you need plus the original error. Just thinking out at loud though.

But, again, maybe `e: unknown` is actually always `e: string`, don't know, was just a curiosity question in case not.

frederikrothenberger: >But, again, maybe e: unknown is actually always e: string, don't know, was just a curiosity question in case not.

That is definitely not the case, because one of the errors we're expecting to catch here is [this one](https://github.com/dignifiedquire/borc/blob/1bd231963344d8ce2dfb9d9bcf04ea29ff4875de/src/decoder.js#L566). Thanks for pointing that out.

@krpeacock: is there a utility for that in `agent-js` somewhere already? Then I'll adjust it.

PR #878, opened 2024-04-29
feat: add support for proof of absence in certificate lookups

nathanosdev:
# Description

Adds support for proof of absence in certificate lookups. Previously a lookup with either return a value or `undefined` if the value cannot be found. Now the result of a lookup will differentiate between `Found`, `Unknown` and `Absent`, which allows clients to tell the difference between a value that is provably absent from the tree, or one that _might_ have been pruned from the tree (maliciously or otherwise).

Fixes # [SDK-1219](https://dfinity.atlassian.net/browse/SDK-1219)

# How Has This Been Tested?

I've added new unit tests to cover the new lookup functionality.

# Checklist:

[SDK-1219]: https://dfinity.atlassian.net/browse/SDK-1219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -29,7 +29,12 @@ import {
 } from './types';
 import { AgentHTTPResponseError } from './errors';
 import { SubnetStatus, request } from '../../canisterStatus';
-import { CertificateVerificationError, HashTree, lookup_path } from '../../certificate';
+import {
+  CertificateVerificationError,
+  HashTree,
+  LookupStatus,
+  lookup_path,
+} from '../../certificate';
 import { ed25519 } from '@noble/curves/ed25519';
 import { ExpirableMap } from '../../utils/expirableMap';
 import { Ed25519PublicKey } from '../../public_key';
@@ -902,14 +907,14 @@ export class HttpAgent implements Agent {
         throw new Error('Could not decode time from response');
       }
       const timeLookup = lookup_path(['time'], tree);
-      if (!timeLookup) {
+      if (timeLookup.status !== LookupStatus.Found) {
         throw new Error('Time was not found in the response or was not in its expected format.');
       }
 
-      if (!(timeLookup instanceof ArrayBuffer) && !ArrayBuffer.isView(timeLookup)) {
+      if (!(timeLookup.value instanceof ArrayBuffer) && !ArrayBuffer.isView(timeLookup)) {
         throw new Error('Time was not found in the response or was not in its expected format.');
       }
-      const date = decodeTime(bufFromBufLike(timeLookup));
+      const date = decodeTime(bufFromBufLike(timeLookup.value as ArrayBuffer));
       this.log('Time from response:', date);
       this.log('Time from response in milliseconds:', Number(date));
       return Number(date);
--- a/packages/agent/src/certificate.test.ts
+++ b/packages/agent/src/certificate.test.ts
@@ -5,7 +5,7 @@
  */
 import * as cbor from './cbor';
 import * as Cert from './certificate';
-import { fromHex, toHex } from './utils/buffer';
+import { bufEquals, fromHex, toHex } from './utils/buffer';
 import { Principal } from '@dfinity/principal';
 import { decodeTime } from './utils/leb';
 import { readFileSync } from 'fs';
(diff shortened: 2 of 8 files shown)


nathanosdev:
@krpeacock it would be good to get your opinion on the interface here. I've kept it so that `Certificate.lookup()` has the same behaviour as before. To get the new lookup behaviour, clients would need to import the `lookup_path()` function.

Would you prefer to keep it like this or to propagate the new behaviour to the `Certificate.lookup()` function too?

It's also possible now to make the agent more robust by retrying requests when requested certificate entries are `Unknown` since that would imply the replica is acting maliciously. Retrying these requests would hit another (hopefully honest) replica and get a certificate that has the expected values. Examples are when we do lookups for `time` or `request_status`, but I think that would be best done in a follow up task.

packages/agent/src/certificate.test.ts
```diff
@@ -20,6 +19,24 @@ function pruned(str: string): ArrayBuffer {
   return fromHex(str);
 }
 
+function bufferEqualityTester(a: unknown, b: unknown): boolean | undefined {
```
krpeacock, line 22: We already have `compare` and `bufEquals` utils in `packages/agent/src/utils/buffer.ts`

krpeacock, line 22: Again, is duplicative of other buffer utils that we already have

packages/agent/src/certificate.ts
```diff
@@ -122,6 +120,17 @@ function isBufferEqual(a: ArrayBuffer, b: ArrayBuffer): boolean {
   return true;
 }
 
+function isBufferGreaterThan(a: ArrayBuffer, b: ArrayBuffer): boolean {
```
krpeacock, line 123: Can use `compare` instead.

`isBufferGreaterThan` is equivalent to `compare(a, b) > 0`

krpeacock:
@nathanosdev I think that extending the changes to the `lookup` and also exporting `lookup_path` is a fine path forward. 

I'd rather not also introduce the retry logic for `unknown` at this stage. It makes me want to think about a different design for retry logic, perhaps offering a more fine-grained set of options giving devs more control over what circumstances the agent will retry under

packages/agent/src/agent/http/index.ts
```diff
@@ -833,14 +838,14 @@ export class HttpAgent implements Agent {
         throw new Error('Could not decode time from response');
       }
       const timeLookup = lookup_path(['time'], tree);
-      if (!timeLookup) {
+      if (timeLookup.status !== LookupStatus.Found) {
         throw new Error('Time was not found in the response or was not in its expected format.');
       }
 
-      if (!(timeLookup instanceof ArrayBuffer) && !ArrayBuffer.isView(timeLookup)) {
+      if (!(timeLookup.value instanceof ArrayBuffer) && !ArrayBuffer.isView(timeLookup)) {
```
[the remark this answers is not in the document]
nathanosdev: `timelookup.value` is either an `ArrayBuffer` or a  `HashTree`. So if we pass `timelookup.value` into `bufFromBufLike` when `timelookup.value` is a `HashTree`, it will pass the `Array.isArray` test in `bufFromBuflike` and be returned as an `ArrayBuffer`. And `decodeTime` may actually succeed, depending on the format of the `HashTree`, and have an incorrect timestamp returned.

It's more robust to exist early if we receive a malformed response in my opinion.

packages/agent/src/certificate.ts
nathanosdev: Nice catch. I'll also remove the `isBufferEqual` function in this file in favor of the common utils

packages/agent/src/certificate.test.ts
nathanosdev: huh... I had understood returning `undefined` was important to tell Jest to use a different equality tester, but this works:
```
(expect as any).addEqualityTesters([bufEquals]);
```

Thanks!

nathanosdev:
> @nathanosdev I think that extending the changes to the `lookup` and also exporting `lookup_path` is a fine path forward.
> 
> I'd rather not also introduce the retry logic for `unknown` at this stage. It makes me want to think about a different design for retry logic, perhaps offering a more fine-grained set of options giving devs more control over what circumstances the agent will retry under

Sounds good, I'll extend the changes to `lookup`.

nathanosdev:
@krpeacock I've extended the changes to the `lookup` method now and used the `lookupResultToBuffer` everywhere that references the `lookup` method to keep the same behaviour as before for internal calls to `lookup`.

There seems to be something up with the `Check PR title` job. There was a new breaking version of `conventional-changelog-angular` a few days ago so I assumed that was the problem, but after locking it down to the previous version that was published several months ago it's still broken. Do you have any idea what might be going on?

krpeacock:
> There seems to be something up with the `Check PR title` job.

Oh, you just have to remove the exclamation mark from the pr title, even though it's proper semantics for conventional commits. Leave it in the change log though

nathanosdev:
It doesn't seem happy with it still, the error message is `Error: Must use import to load ES Module: /action/node_modules/conventional-changelog-angular/src/index.js`

krpeacock (approved):

PR #894, opened 2024-06-14
chore: update management canister interface with wasm_memory_limit

ulan:
This updates the interface of the management canister to include the latest spec change:
https://github.com/dfinity/interface-spec/pull/278


ulan:
Closing because the interface has already been updated

PR #905, opened 2024-07-18
feat: allow for setting HttpAgent ingress expiry using `ingressExpiryInMinutes` option

krpeacock:
# Description

Simply adds a configurable ingress expiry option. Includes checks against <= 0 and > 5 minutes, new errors, and e2e tests about behavior. 

Important question to resolve: when the expiry is less than or equal to one minute, the tests often fail in practice due to the current "round down to the nearest minute". How should the agent handle this? 

> Resolution - now matches the agent-rs strategy and rounds down to the nearest second when the delta is < 90 seconds

My suggestion would be either
- disable expiry rounding if `ingressExpiryInMinutes` is passed
- disable expiry rounding if `ingressExpiryInMinutes` is passed and the value is less than 2 minutes

# How Has This Been Tested?

new e2e tests and unit tests

# Checklist:


krpeacock:
cc @robin-kunzler for prodsec review

dfx-json (approved):

PR #906, opened 2024-07-18
feat: v3 api sync call

krpeacock:
# Description

Implements `sync_call` functionality from the v3 api for improved performance. See https://github.com/dfinity/interface-spec/pull/265 for more details

## Changes:

- feat: sync_call support in HttpAgent and Actor
  - Skips polling if the sync call succeeds and provides a certificate
  - Falls back to v2 api if the v3 endpoint 404's
  - Adds certificate to SubmitResponse endpoint
  - adds callSync option to `HttpAgent.call`, which defaults to `true`

Fixes SDK-1448

# How Has This Been Tested?

Updates to unit tests and e2e tests run against dfx `0.22.0` and `0.23.0` to ensure backwards compatibility

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -464,44 +466,78 @@ export class HttpAgent implements Agent {
...
           ...transformedRequest.request,
           body,
-        }),
-      backoff,
-      tries: 0,
-    });
+        });
+
+      this.log.print(
+        `fetching "/api/v3/canister/${ecid.toText()}/call" with request:`,
```
ericswanson-dfinity, line 487: Looks like this log message is inaccurate if callSync is false

ericswanson-dfinity (changes requested):

packages/agent/src/agent/http/index.ts
krpeacock: Thanks - I've moved the log inside of each request function

ericswanson-dfinity (approved):

packages/agent/src/actor.ts
```diff
@@ -530,25 +530,41 @@ function _createActorMethod(
...
+          lookupResultToBuffer(certificate.lookup([...path, 'status'])),
+        );
+
+        switch (status) {
+          case 'replied':
+            reply = lookupResultToBuffer(certificate.lookup([...path, 'reply']));
+            break;
+          case 'rejected':
+            throw new UpdateCallRejectedError(cid, methodName, requestId, response);
+          case 'accepted':
```
venkkatesh-sekar, line 554: `accepted` is not a valid[ request status](https://internetcomputer.org/docs/current/references/ic-interface-spec#state-tree-request-status) and is not part of the state tree. 

From v3 implementation, the request status could only be either `replied` or `rejected`

packages/agent/src/actor.ts
```diff
@@ -530,25 +530,41 @@ function _createActorMethod(
...
+            // The certificate is not yet ready, so we need to poll for the response
+            break;
+          case 'default':
+            throw new ActorCallError(cid, methodName, 'update', { requestId: toHex(requestId) });
+        }
+      }
+      if (reply === undefined) {
```
venkkatesh-sekar, line 561: We should only poll here if the response.status is 202. Otherwise, we can throw the http error received as the request is guaranteed to be not submitted. 

However, the boundary node can be malicious and send whatever they want, so I'm not quite confident on the right behavior here.

packages/agent/src/actor.ts
```diff
@@ -530,25 +530,41 @@ function _createActorMethod(
...
-      response;
-      requestDetails;
-
-      if (!response.ok || response.body /* IC-1462 */) {
-        throw new UpdateCallRejectedError(cid, methodName, requestId, response);
+      let reply: ArrayBuffer | undefined;
+      let certificate: Certificate | undefined;
+      if (response.body && response.body.certificate) {
+        const cert = response.body.certificate;
+        certificate = await Certificate.create({
```
venkkatesh-sekar, line 537: We can check if agent.rootKey is initialized before verifying as done in `pollForResponse`

packages/agent/src/actor.ts
venkkatesh-sekar: cc: @robin-kunzler

krpeacock: I disagree with this interpretation actually - reply is undefined when it is defined on line 533, and I have set up this handler to handle both the `v2` and `v3` behavior.

`reply` remains undefined if there is no `body` in the response, which is the v2 behavior and is only set if we are able to parse a certificate in the response. 

The Actor is unaware of which strategy the `agent` is using. This is a deliberate choice to maintain compatibility with earlier agents, which are still very popular in use with the system. If a wallet or provider hasn't upgraded, this design will allow the Actor to use an agent that only hits the `v2` endpoint, correctly following up by polling for response.

I can make this clearer in comments, variable names, or so on, however

venkkatesh-sekar: Ah, I missed the v2 fallback but the situation shouldn't change. Even [under v2](https://github.com/dfinity/ic/blob/d436a526d27f528d093f75161a6382c8d2425cc3/rs/http_endpoints/public/src/call/call_v2.rs#L131), `202 ACCEPTED` is the status code from the replica to know that the ingress has been submitted

krpeacock: Good to know, I'll make the check more specific in that case!

packages/agent/src/actor.ts
```diff
@@ -525,35 +525,51 @@ function _createActorMethod(
...
+          lookupResultToBuffer(certificate.lookup([...path, 'status'])),
+        );
+
+        switch (status) {
+          case 'replied':
+            reply = lookupResultToBuffer(certificate.lookup([...path, 'reply']));
+            break;
+          case 'rejected':
+            throw new UpdateCallRejectedError(cid, methodName, requestId, response);
+          case 'default':
```
venkkatesh-sekar, line 557: Just for my understanding, is this equivalent to adding a `default:` to a `switch` stmt? In that case, we wouldn't want it to throw but go into the polling loop. 

Maybe making it more explicit would make sense? 
- received, processing -> break and poll for response 
- done or default -> throw ActorCallError

venkkatesh-sekar (approved):

packages/agent/src/actor.ts
krpeacock: It is not the same as a default case - I wrote this originally a few weeks ago, and can't recall why this was here. It can simply be removed and the logic below will apply

Max-SDSharp:
@krpeacock 
Only call calls that access the V3 API?
read_state and query continue with the V2 API or is it possible to access V3 as well?

PR #940, opened 2024-10-11
fix: trap and throw handling in v3 sync call

krpeacock:
# Description

Reported after observing that `reject_code` and `reject_message` were missing in the error message in cases where v3 sync calls were used. Corrects the flow to pull this information out of the certificate and provides them to the error

Also adds them as public fields on the error object

Fixes # (issue)

# How Has This Been Tested?

New e2e tests handling `Trap` and `Throw` cases

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -414,8 +415,9 @@ export class HttpAgent implements Agent {
     },
     identity?: Identity | Promise<Identity>,
   ): Promise<SubmitResponse> {
+    // TODO - restore this value
     const callSync = options.callSync ?? true;
-    const id = await (identity !== undefined ? await identity : await this.#identity);
+    const id = await(identity !== undefined ? await identity : await this.#identity);
```
dfx-json, line 420: You only need the outer await here

dfx-json (approved):

PR #941, opened 2024-10-11
feat: call and poll

krpeacock:
# Description

Adds a new utility - `callAndPoll`. This function allows you to make a call using the `HttpAgent` and retrieving the raw `certificate`, either using the v3 sync call or falling back to polling. This duplicates the behavior done by `Actor` at a low level, and is intended to be used by other libraries.

Fixes # SDK-1828

# How Has This Been Tested?

New e2e test

# Checklist:


packages/agent/src/utils/callAndPoll.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function callAndPoll(options: {
+  canisterId: Principal | string;
+  methodName: string;
+  agent: Agent;
+  arg: ArrayBuffer;
+}): Promise<ArrayBuffer> {
+  const { canisterId, methodName, agent, arg } = options;
+  const cid = Principal.from(options.canisterId);
+
+  const { defaultStrategy } = polling.strategy;
```
dfx-json, line 23: Would it be useful for polling strategy to be an option?

dfx-json (approved):

packages/agent/src/utils/callAndPoll.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param options.agent - The agent to use to make the call
+ * @param options.arg - The argument to pass to the canister
+ * @returns The certificate response from the canister (which includes the reply)
+ */
+export async function callAndPoll(options: {
+  canisterId: Principal | string;
+  methodName: string;
+  agent: Agent;
+  arg: ArrayBuffer;
+}): Promise<ArrayBuffer> {
```
frederikrothenberger, line 19: We need to get back the `contentMap` _and_ the `rawCert`. Ideally, this function would have the exact same interface as `call`, maybe with the addition of being able to supply the `pollStrategy` as suggested by @dfx-json.

Also, please extend the `CallOptions` interface to allow specifying the `nonce` too.

krpeacock: @frederikrothenberger can you define what `contentMap` precisely is? Should it be an ArrayBuffer? Does it need to be decoded into a JavaScript object? Are you willing to provide an IDL interface for the types so we can actually decode the candid?

Relatedly, I don't know what actual feature this is for, or how it will be used in practice. We'd probably have had a lot less back-and-forth about this if I was just designing a feature for your use-case instead of all these half-requirements and us each having slightly different definitions of what things are

e2e/node/basic/callAndPoll.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    const options = {
+      canisterId: 'tnnnb-2yaaa-aaaab-qaiiq-cai',
+      methodName: 'inc_read',
+      agent: await HttpAgent.create({ host: 'https://icp-api.io' }),
+      arg: fromHex('4449444c0000'),
+    };
+
+    const certificate = await callAndPoll(options);
+    expect(certificate instanceof ArrayBuffer).toBe(true);
```
peterpeterparker, line 15: Are you going to enhance the tests? e.g. asserting that the certificate return is the expected value or asserting that the function correctly throws if there was an error?

e2e/node/basic/trap.test.ts
```diff
@@ -4,7 +4,14 @@ import util from 'util';
 import exec from 'child_process';
 const execAsync = util.promisify(exec.exec);
 
-const { stdout } = await execAsync('dfx canister id trap');
+// eslint-disable-next-line prefer-const
+let stdout;
```
peterpeterparker, line 8: Shouldn't this be provided in another PR?

packages/agent/src/utils/callAndPoll.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (agent.rootKey == null) throw new Error('Agent root key not initialized before making call');
+
+  const { requestId, response } = await agent.call(cid, {
+    methodName,
+    arg,
+    effectiveCanisterId: cid,
+  });
+
+  let certificate: Certificate;
+  if (response.body && (response.body as v3ResponseBody).certificate) {
```
peterpeterparker, line 34: Should you assert that the type is v3 before casting? This feels a bit optimistic.

peterpeterparker, line 34: While a response with a `body.certificate` is valid only for status 200 according specification I would personnaly still validate the status anyway

packages/agent/src/utils/callAndPoll.ts
```diff
@@ -0,0 +1,54 @@
...
+    const cert = (response.body as v3ResponseBody).certificate;
+    // Create certificate to validate the responses
+    certificate = await Certificate.create({
```
peterpeterparker, line 37: Shouldn't the certificate `status` be asserted at this point?

packages/agent/src/utils/callAndPoll.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+  // Fall back to polling if we recieve an Accepted response code
+  if (response.status === 202) {
+    const pollStrategy = defaultStrategy();
+    // Contains the certificate and the reply from the boundary node
+    const response = await polling.pollForResponse(agent, cid, requestId, pollStrategy);
+    certificate = response.certificate;
+  }
+
+  return certificate.rawCert;
```
peterpeterparker, line 53: Shouldn't it throws if none of the above was a valid response?

peterpeterparker (commented):
A few comments in addition to those of @frederikrothenberger  by comparing with what we have implemented in the Oisy Wallet Signer library.

Source: https://github.com/dfinity/oisy-wallet-signer/blob/main/src/agent/custom-http-agent.ts

e2e/node/basic/callAndPoll.test.ts
krpeacock: Yes, this should have been marked as a draft. I wanted feedback on the inputs and return values before proceeding. If you have real-world test cases this feature will be used for, those would be great to add as tests as well, beyond the basic cases!

packages/agent/src/utils/callAndPoll.ts
krpeacock: I wasn't (and still won't be until tomorrow) sure what the intended purpose for this feature is. `Certificate.create` will validate the certificate, but not check the status. There's no problem with throwing if it's rejected, but I wasn't sure if that was desireable behavior for this low-level feature

packages/agent/src/utils/callAndPoll.ts
krpeacock: fair enough

PR #945, opened 2024-10-17
chore: skip range check when the certificate comes from the management canister

dfx-json:
Resolves SDKTG-403 and a regression in the NNS dapp in versions of agent > 2.0.0


krpeacock:
I think this is safe to add. We were overriding this check with a custom transformation for our `getManagementCanister` utility anyway, and this will make the tools more reliable for downstream features

krpeacock:
@dfx-json did you try this with any live code? Do we know that `@dfinity/ic-management` was passing the correct <canister-id> vs the effective-canister-id, such that this check would match against `aaaaa-aa`?

krpeacock:
image

Tested using @dfinity/ic-management - I can still reproduce the error with these changes

krpeacock (approved):
Looks good to go, and I'm happy to have this new unit testing pattern. It's something we can probably polish and automate down the line, and will be improved with nicer tooling for passing fixed nonces

Issue #773, opened 2023-09-26
BLS-verify bundling issue


Gekctek:
I am trying to integrate the bls-verify package into a raw js file for some WebGL unity project for ICP.NET
The issue I am having is that I am doing `npx esbuild --bundle src/index.ts --outfile=dist/index.js --platform=browser` to bundle up the package to include as raw js, but the issue is that it breaks because all the files in `packages/bls-verify/src/vendor/amcl-js/src` are not included except for ctx.js. It seems to be an issue because of the dynamic loading of the other files like aes.js. When it tries to load them it breaks
How do you guys handle this?

krpeacock:
Sorry for the inconvenience - the BLS verify feature is going to be deprecated and replaced soon with a native JS verification strategy from `@noble/curves`

PR #954, opened 2024-11-26
feat: effective target canister ID for mgmt call

peterpeterparker:
# Description

Similarly to ic-js (see PR [#773](https://github.com/dfinity/ic-js/pull/773)), the goal of this PR is to extend the `transform` function of the `getManagementCanister` to support calls on the mainnet to `install_chunked_code`. While I have not tested the function of agent-js irl, I can confirm that without extending the similar function in `ic-js`, I am unable to perform such calls on the mainnet.

This issue is due to the following specification:

> If the request is an update call to the Management Canister (`aaaaa-aa`), then:
> 
> If the call is to the `install_chunked_code` method and the `arg` is a Candid-encoded record with a `target_canister` field of type `principal`, then the effective canister id must be that principal.

See [documentation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-effective-canister-id)

# Notes

Because I am not familiar with the agent-js codebase and want to avoid introducing too many changes, this PR solely focuses on adding the necessary code for the intended functionality. There are likely ways to improve the type declarations and address some code duplication after this implementation, but I did not want to add unnecessary noise.

There is also a potential "race condition" if a `transform` function is called with both `canister_id` and `target_canister`. In such cases, `target_canister` will take precedence over `canister_id`. For the same reasons mentioned above, I did not want to add additional complexity to the implementation, as this was not the primary goal of the PR. However, it is likely an issue worth addressing in the future.

Similarly, this PR includes a test specifically for the `target_canister` mapping to `effective_canister_id`, which duplicates some existing mocking. It would likely be beneficial to reduce code duplication in this area. Additionally, it would be useful to add tests for the other mappings of `getManagementCanister`, but this is beyond the scope of this feature.


krpeacock (changes requested):
Looks good! Can you add a line to the Changelog before we merge?

peterpeterparker:
Sure. I added a line to the CHANGELOG.

krpeacock (approved):

PR #959, opened 2024-12-07
build: fix @peculiar/webcrypto is no peer dependencies but dev dependencies

peterpeterparker:
# Description

`@peculiar/webcrypto` is not a peer dependency of `@dfinity/identity` but a dev dependency.


krpeacock (approved):

krpeacock:
Yeah, we need another SDK member to approve since I pushed the last change

ericswanson-dfinity (approved):

PR #960, opened 2024-12-11
chore: pins nanoid dev dependency version to override warning

krpeacock:
also removes rollup pin which was added a while ago for similar reasons and is no longer needed


lwshang (dismissed):

krpeacock:
Note - the size limit seems to have decreased a lot by updating the esbuild version! Unrelated to code changes, but end users will probably benefit from newer versions of tools that depend on it

lwshang (approved):

PR #963, opened 2025-01-13
feat: await fetchRootKey calls while making API calls

nathanosdev:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -150,6 +150,14 @@ export interface HttpAgentOptions {
    * Alternate root key to use for verifying certificates. If not provided, the default IC root key will be used.
    */
   rootKey?: ArrayBuffer;
+
+  /**
+   * Whether the agent should fetch the root key from the network. Defaults to false.
+   *
+   * WARNING!!! Do not enable this in production environments,
+   * as it can be used as an attack vector by malicious nodes.
+   */
+  shouldFetchRootKey?: boolean;
 }
 
 function getDefaultFetch(): typeof fetch {
@@ -245,7 +253,6 @@ other computations so that this class can stay as simple as possible while
 allowing extensions.
  */
 export class HttpAgent implements Agent {
-  public rootKey: ArrayBuffer;
   #identity: Promise<Identity> | null;
   readonly #fetch: typeof fetch;
   readonly #fetchOptions?: Record<string, unknown>;
@@ -253,10 +260,13 @@ export class HttpAgent implements Agent {
   #timeDiffMsecs = 0;
   readonly host: URL;
   readonly #credentials: string | undefined;
-  #rootKeyFetched = false;
   readonly #retryTimes; // Retry requests N times before erroring by default
   #backoffStrategy: BackoffStrategyFactory;
   readonly #maxIngressExpiryInMinutes: number;
+
+  #rootKey: ArrayBuffer;
+  readonly #shouldFetchRootKey: boolean;
+  #fetchRootKeyPromise: Promise<ArrayBuffer> | null = null;
 
   // Public signature to help with type checking.
   public readonly _isAgent = true;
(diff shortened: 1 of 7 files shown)


packages/agent/src/agent/http/index.ts
```diff
@@ -288,7 +298,11 @@ export class HttpAgent implements Agent {
     this.#fetch = options.fetch || getDefaultFetch() || fetch.bind(global);
     this.#fetchOptions = options.fetchOptions;
     this.#callOptions = options.callOptions;
-    this.rootKey = options.rootKey ? options.rootKey : fromHex(IC_ROOT_KEY);
+
+    this.#rootKey = options.rootKey ?? fromHex(IC_ROOT_KEY);
+    this.#shouldFetchRootKey = options.shouldFetchRootKey ?? false;
+    // kick off the fetchRootKey process asynchronously, if needed
+    (async () => await this.fetchRootKey())();
```
nathanosdev, line 305: We start fetching the root key as soon as possible, in practice this should mean that there are no further delays compared to creating the agent with an async constructor and waiting for this task to complete before making API calls.

packages/agent/src/agent/http/index.ts
```diff
@@ -245,19 +253,21 @@ other computations so that this class can stay as simple as possible while
 allowing extensions.
  */
 export class HttpAgent implements Agent {
-  public rootKey: ArrayBuffer;
```
nathanosdev, line 248: I've removed the public `rootKey` property in favor of the async `getRootKey`, since we cannot guarantee availability of the correct root key when `fetchRootKey` is set to true.

packages/agent/src/agent/http/index.ts
```diff
@@ -1125,14 +1132,27 @@ export class HttpAgent implements Agent {
...
+
   public async fetchRootKey(): Promise<ArrayBuffer> {
-    if (!this.#rootKeyFetched) {
-      const status = await this.status();
-      // Hex-encoded version of the replica root key
-      this.rootKey = (status as JsonObject & { root_key: ArrayBuffer }).root_key;
-      this.#rootKeyFetched = true;
+    if (!this.#shouldFetchRootKey) {
+      return this.#rootKey;
+    }
```
nathanosdev, line 1142: If we should not fetch the root key, we return the existing root key, which is either a hard coded root key provided by the consumer or the mainnet root key.

packages/agent/src/agent/http/index.ts
```diff
@@ -1125,14 +1132,27 @@ export class HttpAgent implements Agent {
...
+
+    if (this.#fetchRootKeyPromise === null) {
+      this.#fetchRootKeyPromise = this.#fetchRootKey();
     }
```
nathanosdev, line 1146: If we should fetch the root key, and this is the first time this function is called, we create the promise to fetch the root key.

packages/agent/src/agent/http/index.ts
```diff
@@ -1125,14 +1132,27 @@ export class HttpAgent implements Agent {
...
-      const status = await this.status();
-      // Hex-encoded version of the replica root key
-      this.rootKey = (status as JsonObject & { root_key: ArrayBuffer }).root_key;
-      this.#rootKeyFetched = true;
+    if (!this.#shouldFetchRootKey) {
+      return this.#rootKey;
+    }
+
+    if (this.#fetchRootKeyPromise === null) {
+      this.#fetchRootKeyPromise = this.#fetchRootKey();
```
nathanosdev, line 1145: If we should fetch the root key, and this is the first time this function is called, we create the promise to fetch the root key.

packages/agent/src/agent/http/index.ts
```diff
@@ -1125,14 +1132,27 @@ export class HttpAgent implements Agent {
...
     }
-    return this.rootKey;
+
+    return await this.#fetchRootKeyPromise;
```
nathanosdev, line 1148: Finally we await the promise and return the resulting root key.

packages/agent/src/agent/http/index.ts
nathanosdev: Agreed that functionally the two approaches are the same.

Doing it this way was a stylistic choice to make it very explicit to any reader of the code that we're invoking a promise but not awaiting it on purpose. I'm fine with just invoking the promise.

krpeacock:
closed in favor of the solution in #966

PR #966, opened 2025-01-25
feat: fetch root key before making calls

krpeacock:
# Description
Since the root key cannot be fetched inside a synchronous constructor, we should await the root key when the `shouldFetchRootKey` flag is set.

This PR adds a guard to the `HttpAgent` async calls to make sure the root key call has finished before making the call. It also modifies some checks for `rootKey` in the `canisterStatus` and `actor` utils that should instead wait for the call to be completed.

This shores up occasional inconsistencies with the agent during local development, and that have affected projects in our examples repostory

Fixes # SDK-1905

# How Has This Been Tested?

New tests added to the HttpAgent unit test suite

# Checklist:

### DIFF
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -1146,19 +1146,27 @@ export class HttpAgent implements Agent {
   }
 
   public async fetchRootKey(): Promise<ArrayBuffer> {
-    // Wait for already pending promise
+    let result: ArrayBuffer;
+    // Wait for already pending promise to avoid duplicate calls
     if (this.#rootKeyPromise) {
-      const result = await this.#rootKeyPromise;
-      this.#rootKeyPromise = null;
-      this.rootKey = result;
-      return result;
+      result = await this.#rootKeyPromise;
+    } else {
+      // construct promise
+      this.#rootKeyPromise = new Promise<ArrayBuffer>((resolve, reject) => {
+        this.status()
+          .then(value => {
+            // Hex-encoded version of the replica root key
+            const rootKey = (value as JsonObject & { root_key: ArrayBuffer }).root_key;
+            this.rootKey = rootKey;
+            resolve(rootKey);
+          })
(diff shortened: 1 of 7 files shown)


krpeacock:
There is one arguably breaking change - the `rootKey` is now typed as `ArrayBuffer | null`. However, I'd say that this typing makes the reality of the workflow explicit. In the past, we would always set the `rootKey` to the `IC_ROOT_KEY` regardless of whether we were going to fetch one or not, leading to the potential asynchronous errors we've already encountered. 

This does mean that according to TypeScript, the `HttpAgent` exported in this version wouldn't be usable in an `Actor` from version `2.0.0`. However, we generally see the reverse pattern in practice. Generally people use an up-to-date `Actor` with an outdated agent.

I think this can be shipped as a minor version update, for this reason.

packages/agent/src/agent/http/index.ts
```diff
@@ -1126,13 +1146,31 @@ export class HttpAgent implements Agent {
   }
 
   public async fetchRootKey(): Promise<ArrayBuffer> {
```
nathanosdev, line 1148: `#rootKeyPromise` is never actually set so the .

```typescript
  public async fetchRootKey(): Promise<ArrayBuffer> {
    if (this.#fetchRootKeyPromise === null) {
      this.#fetchRootKeyPromise = this.#fetchRootKey();
    }

    const result = await this.#fetchRootKeyPromise;
    this.rootKey = result;
    return result;
  }

  async #fetchRootKey(): Promise<ArrayBuffer> {
    const status = await this.status();
    // Hex-encoded version of the replica root key
    const rootKey = (status as JsonObject & { root_key: ArrayBuffer }).root_key;
    this.rootKey = rootKey;

    return rootKey;
  }
```

krpeacock: Thanks for catching!

ericswanson-dfinity (approved):

PR #979, opened 2025-03-17
fix: add backtracking to opt decoding

crusso:
# Description

Fixes decoding of variants under option types, allowing compatible tagged values from wider options, defaulting incompatible tagged values to null.

Brings decoding of options closer to spec, but see child PR #981 for the remaining edge cases.

This PR is marked `do not merge` until #981 is merged into this one.

# How Has This Been Tested?

See PR for additional candid tests.

# Checklist:

What CHANGELOG?

### DIFF
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -717,12 +717,12 @@ test('IDL opt variant decoding', () => {
     IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null }))}),
     {a: [{ x: null }]},
     '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
-    'extended variant under opt existing',
+    'extended variant under opt expected tag',
   );
   testDecode(
     IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null }))}),
     {a: []},
     '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
-    'extended variant under opt new - defaulting',
+    'extended variant under opt unexpected tag - defaulting',
   );
 });
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -698,3 +698,31 @@ test('should decode matching optional fields if wire type contains additional fi
     b: ['123'],
   });
 });
+
+
+test('IDL opt variant decoding', () => {
+  testDecode(
+    IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null, z: IDL.Null }))}),
+    {a: [{ x: null }]},
+    '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
+    'same variant under opt x',
+  );
+  testDecode(
+    IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null, z: IDL.Null }))}),
+    {a: [{ z: null}]},
+    '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
+    'same variant under opt z',
+  );
+  testDecode(
+    IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null }))}),
+    {a: [{ x: null }]},
+    '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
+    'extended variant under opt existing',
+  );
+  testDecode(
+    IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null }))}),
+    {a: []},
+    '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
+    'extended variant under opt new - defaulting',
+  );
+});
--- a/packages/candid/src/idl.test.ts
+++ b/packages/candid/src/idl.test.ts
@@ -703,7 +703,7 @@ test('should decode matching optional fields if wire type contains additional fi
 describe('IDL opt variant decoding', () => {
   it('should handle the first of three variants', () => {
     testDecode(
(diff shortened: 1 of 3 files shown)


packages/candid/src/idl.ts
```diff
@@ -921,8 +921,19 @@ export class OptClass<T> extends ConstructType<[T] | []> {
...
-        return [this._type.decodeValue(b, opt._type)];
+      case 1: {
+        const checkpoint = b.save();
+        try {
+          const v = this._type.decodeValue(b, opt._type)
+          return [v];
+        } catch (e : any) {
+          b.restore(checkpoint);
+          // skip value at wire type (to advance b)
+          const v = opt._type.decodeValue(b, opt._type)
```
krpeacock, line 932: Why do we decode the value again if we are always returning none?

crusso: Because we need to skip past the value so the buffer is consumed...

crusso: The motoko code does this a bit differently. It avoids reparsing the candid by parsing to completion and returning a proper value or sentinel value to indicate coercion failure. But that's harder to integrate here without introducing mistakes.

packages/candid/src/idl.test.ts
```diff
@@ -698,3 +698,31 @@ test('should decode matching optional fields if wire type contains additional fi
...
+    '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
+    'extended variant under opt expected tag',
+  );
+  testDecode(
+    IDL.Record({a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null }))}),
+    {a: []},
+    '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
+    'extended variant under opt unexpected tag - defaulting',
+  );
+});
```
krpeacock, line 728: Some basic reformatting so the individual failures will actually be labeled if the tests failed
```suggestion
describe('IDL opt variant decoding', () => {
  it('should handle the first of three variants', () => {
    testDecode(
      Record({ a: Opt(Variant({ x: Null, y: Null, z: Null })) }),
      { a: [{ x: null }] },
      '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}},
      'same variant under opt x',
    );
  });
  it('should handle the third of three variants', () => {
    testDecode(
      Record({ a: Opt(Variant({ x: Null, y: Null, z: Null })) }),
      { a: [{ z: null }] },
      '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
      'same variant under opt z',
    );
  });
  it('should handle the first of two variants', () => {
    testDecode(
      Record({ a: Opt(Variant({ x: Null, y: Null })) }),
      { a: [{ x: null }] },
      '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
      'extended variant under opt expected tag',
    );
  });
  it('should handle the option when the option is empty', () => {
    testDecode(
      Record({ a: Opt(Variant({ x: Null, y: Null })) }),
      { a: [] },
      '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
      'extended variant under opt unexpected tag - defaulting',
    );
  });
});
```

packages/candid/src/idl.ts
```diff
@@ -921,8 +921,19 @@ export class OptClass<T> extends ConstructType<[T] | []> {
...
+          // retun none
+          return [];
+        }
+      }
```
krpeacock, line 936: More comments for small-brain people like me trying to understand what's happening
```suggestion
      case 1: {
        // Save the current state of the Pipe `b` to allow rollback in case of an error
        const checkpoint = b.save();
        try {
          // Attempt to decode a value using the `_type` of the current instance
          const v = this._type.decodeValue(b, opt._type);
          // If decoding succeeds, return the value wrapped in an array
          return [v];
        } catch (e: any) {
          // If an error occurs during decoding, restore the Pipe `b` to its previous state
          b.restore(checkpoint);
          // Skip the value at the current wire type to advance the Pipe `b` position
          opt._type.decodeValue(b, opt._type);
          // Return an empty array to indicate a `none` value
          return [];
        }
      }
```

packages/candid/src/idl.test.ts
```diff
@@ -698,3 +698,39 @@ test('should decode matching optional fields if wire type contains additional fi
     b: ['123'],
   });
 });
+
+
+describe('IDL opt variant decoding', () => {
```
jasonz-dfinity, line 703: It looks like the first 3 test cases can be tested for round trip (use `test_` instead of `testDecode`)?

packages/candid/src/idl.test.ts
```diff
@@ -698,3 +698,39 @@ test('should decode matching optional fields if wire type contains additional fi
...
+      { a: [{ x: null }] },
+      '4449444c036c0161016e026b03787f797f7a7f01000100', // Motoko: {a = ?#x} : {a : ?{#x;#y;#z}}
+      'extended variant under opt expected tag',
+    );
+  });
+  it('should handle the option when the option is empty', () => {
+    testDecode(
+      IDL.Record({ a: IDL.Opt(IDL.Variant({ x: IDL.Null, y: IDL.Null })) }),
+      { a: [] },
+      '4449444c036c0161016e026b03787f797f7a7f01000102', // Motoko: {a = ?#z} : {a : ?{#x;#y;#z}}
```
jasonz-dfinity, line 732: IMHO if you put this case right after the second test case, it would be more easily understood without knowledge of motoko (as they have the same binary format), assuming (most?) humans can't parse candid binary 😄

packages/candid/src/idl.test.ts
crusso: Tried that but they don't round-trip because the encoding was produced by motoko. There is no unique encoding in Candid so this would actually be too much to expect, I think.

packages/candid/src/idl.test.ts
crusso: You can't read the binary?

crusso: I was organizing these cases differently - parsing value encoded at expected type and parsing values encoded at  subtype of expected type, with expected tag or unexpected tag.

crusso: Anyway, I hope you don't mind but I'm closing this PR in favor of the more comprehensive child PR #981 (which has more tests too).

crusso:
Closing in favor of #981 (which incorporates these changes)

PR #985, opened 2025-03-26
fix: make isAuthenticated validate delegation expiry

asabates-dfinity:
# Description

Currently `isAuthenticated` does not work if the delegation has expired, as it does not check at any moment further than the chain has being unset, which only happens when actively logout call is triggered. This produces the scenario in which a consumer can call to `isAuthenticated`, get a true result and yet their agents are rejected because of the expired authorization. 

This PR aims to fix this problem by adding a check validating the delegation chain, it also adjusts the signature of the method to non asynchronous as none of the operations done inside are, indeed, asynchronous. This simplifies the usage for the consumer.

# How Has This Been Tested?

Updated unit tests to check that it returns false when its expired

# Checklist:


packages/auth-client/src/index.ts
```diff
@@ -428,8 +434,12 @@ export class AuthClient {
     return this._identity;
   }
 
-  public async isAuthenticated(): Promise<boolean> {
-    return !this.getIdentity().getPrincipal().isAnonymous() && this._chain !== null;
+  public isAuthenticated(): boolean {
```
krpeacock, line 437: I prefer to leave this asynchronous. If we decide to add support for validating the signatures for all delegations in the chain (currently haven't adopted this yet because we wanted to avoid the secp256k1 dependency), this operation will most likely have async behavior.

Breaking the API is costly, and this is why we tagged it async preemptively, even years in advance

packages/use-auth-client/src/use-auth-client.ts
```diff
@@ -81,10 +81,10 @@ export function useAuthClient(options?: UseAuthClientOptions) {
             logout();
           }),
       },
-    }).then(async client => {
+    }).then(client => {
       setAuthClient(client);
       setIdentity(client.getIdentity());
-      setIsAuthenticated(await client.isAuthenticated());
```
krpeacock, line 87: see here also, I'd like this reverted

docs/CHANGELOG.md
krpeacock, line 1: This is a large formatting diff, presumably from autoformatting. Which tool did you use for this?

krpeacock (changes requested):

packages/auth-client/src/index.ts
asabates-dfinity: Makes sense, reverted here https://github.com/dfinity/agent-js/pull/985/commits/fb339a1f3cf5664a2502fa535ae64298c10efb01

packages/use-auth-client/src/use-auth-client.ts
asabates-dfinity: Fixed here https://github.com/dfinity/agent-js/pull/985/commits/fb339a1f3cf5664a2502fa535ae64298c10efb01

docs/CHANGELOG.md
asabates-dfinity: Fixed here https://github.com/dfinity/agent-js/pull/985/commits/fb339a1f3cf5664a2502fa535ae64298c10efb01, I was just using VSCode with prettier and eslint :)

krpeacock (approved):

PR #981, opened 2025-03-17
fix: make IDL/Candid decoding of options spec compliant

crusso:
# Description

Bring Candid decoding of `opt` types up to Candid spec:

In particular, when decoding at an `opt` type:
  - If the wire type is an `opt` type, decode its payload at the expected content type
    (as before).
  - Allow decoding `null` wire type as IDL value `null` (i.e. JS `[]`).
  - Allow decoding of value of `reserved` wire type, defaulting to IDL value `null` (i.e. JS `[]`).
  - Allow decoding of wider variant type on the wire at narrower expected variant type,
    provided the decoded value is valid at the expected variant type. Otherwise, default to `null` (i.e. JS `[]`).
  - Otherwise:
     - If the expected content type is `null` or `reserved` or (nested) `opt`, return IDL value `null` (i.e. JS `[]`).
     - The expected content type is neither `null`, `reserved` or nested `opt`:
       allow decoding of the non-optioned value `v` as `opt v` (JS `[v*]`) if compatible with
       the expected content type; if incompatible, return IDL value `null` (JS `[]`).

# How Has This Been Tested?

See additional candid tests in PR.

# Checklist:

- [x] My changes follow the guidelines in [CONTRIBUTING.md](https://github.com/dfinity/agent-js/blob/main/CONTRIBUTING.md).
- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I have edited the CHANGELOG accordingly.
- [x] I have made corresponding changes to the documentation (no real change - c.f.  current candid spec)

PR #994, opened 2025-04-10
fix: checks subtyping relation when decoding reference types in Candid

christoph-dfinity:
# Description

Implements Candid subtyping and applies it when decoding reference types, to get up to par with the current Candid spec.

At a user-level this means we will fail early when decoding something like:

`service { foo : (text) -> () }` at an expected type of `service { foo : (nat) -> () }`

rather than decoding successfully and only failing when actually trying to make the call to `foo` later on in the code.

Follow-up to https://github.com/dfinity/agent-js/pull/981

# How Has This Been Tested?

Additional unit tests

I also had to update one of the existing tests, as the binary value it was checking against actually had the wrong type

# Reviewer notes

Unfortunately there are a few formatting changes unrelated to the change itself. These were produced in 3bedee41f13a9027e79c4420fd379d1cd35be18c by running 
```
npm run -w packages/candid prettier:write
```
I added some tests, but it would probably be a good idea to write a parser for the test format in the spec and make sure this implementation conforms to it. I'd prefer doing that in another PR though. Up to you, dear reviewer, if we'd rather hold off on this change until then.

I added a global subtype cache that is reset before decoding a Candid message. As it's global and mutable, make sure to take a closer look if I messed up...

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,7 @@
         "publicOnly": true
       }
     ],
+    "jsdoc/check-tag-names": ["warn", { "definedTags": ["jest-environment"] }],
     "jsdoc/require-returns": "off",
     "jsdoc/require-returns-description": "warn"
   }
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,4 +26,4 @@ jobs:
           node-version: ${{ matrix.node }}
       - run: npm install -g npm
       - run: npm install
-      - run: npm run lint --workspaces --if-present -- --max-warnings=0
+      - run: npm run lint
--- a/e2e/node/package.json
+++ b/e2e/node/package.json
@@ -6,12 +6,10 @@
   "scripts": {
     "ci": "npm run e2e",
     "e2e": "vitest",
-    "eslint:fix": "npm run lint -- --fix",
-    "eslint": "eslint --ext '.js,.jsx,.ts,.tsx' basic canisters utils *.ts",
-    "lint": "npm run eslint",
+    "lint:fix": "npm run lint -- --fix",
+    "lint": "eslint --ext '.js,.jsx,.ts,.tsx' basic canisters utils *.ts",
     "mitm": "MITM=true vitest ./basic/mitm.test.ts",
     "build": "",
-    "lint:fix": "npm run lint -- --fix",
     "test:coverage": "",
     "test": ""
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -18349,6 +18349,132 @@
         "@scure/bip39": "^1.3.0",
         "asn1js": "^3.0.5",
         "bs58check": "^3.0.1"
+      },
+      "devDependencies": {
+        "eslint": "^8.19.0"
+      }
+    },
+    "packages/identity-secp256k1/node_modules/@eslint/js": {
+      "version": "8.57.1",
+      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-8.57.1.tgz",
+      "integrity": "sha512-d9zaMRSTIKDLhctzH12MtXvJKSSUhaHcjV+2Z+GK+EEY7XKpP5yR4x+N3TAcHTcu963nIr+TMcCb4DBCYX1z6Q==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+      }
+    },
+    "packages/identity-secp256k1/node_modules/@humanwhocodes/config-array": {
+      "version": "0.13.0",
+      "resolved": "https://registry.npmjs.org/@humanwhocodes/config-array/-/config-array-0.13.0.tgz",
+      "integrity": "sha512-DZLEEqFWQFiyK6h5YIeynKx7JlvCYWL0cImfSRXZ9l4Sg2efkFGTuFf6vzXjK1cq6IYkU+Eg/JizXw+TD2vRNw==",
+      "deprecated": "Use @eslint/config-array instead",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@humanwhocodes/object-schema": "^2.0.3",
+        "debug": "^4.3.1",
+        "minimatch": "^3.0.5"
+      },
+      "engines": {
+        "node": ">=10.10.0"
+      }
+    },
+    "packages/identity-secp256k1/node_modules/eslint": {
+      "version": "8.57.1",
+      "resolved": "https://registry.npmjs.org/eslint/-/eslint-8.57.1.tgz",
+      "integrity": "sha512-ypowyDxpVSYpkXr9WPv2PAZCtNip1Mv5KTW0SCurXv/9iOpcrH9PaqUElksqEB6pChqHGDRCFTyrZlGhnLNGiA==",
+      "deprecated": "This version is no longer supported. Please see https://eslint.org/version-support for other options.",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@eslint-community/eslint-utils": "^4.2.0",
+        "@eslint-community/regexpp": "^4.6.1",
+        "@eslint/eslintrc": "^2.1.4",
+        "@eslint/js": "8.57.1",
+        "@humanwhocodes/config-array": "^0.13.0",
+        "@humanwhocodes/module-importer": "^1.0.1",
+        "@nodelib/fs.walk": "^1.2.8",
+        "@ungap/structured-clone": "^1.2.0",
+        "ajv": "^6.12.4",
+        "chalk": "^4.0.0",
+        "cross-spawn": "^7.0.2",
+        "debug": "^4.3.2",
+        "doctrine": "^3.0.0",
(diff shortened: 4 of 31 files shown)


docs/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
 - fix: fixes a bug in the Ed25519KeyIdentity verify implementation where the argument order was incorrect
 - fix: fixes a bug in the `Principal` library where the management canister id util was incorrectly importing using `fromHex`
 - feat: change auth-client's default identity provider url
+- fix: Check subtyping relationship when decoding function or service references
```
krpeacock, line 20: can you move this up to the unreleased section?

krpeacock (approved):
one comment on fixing the changelog, but approved

docs/CHANGELOG.md
christoph-dfinity: Done, thanks for noticing.

packages/candid/src/idl.test.ts
```diff
@@ -928,3 +933,191 @@ describe('IDL opt edge cases (embedded)', () => {
...
+  });
+
+  describe('Subtyping on Vecs', () => {
+    testReflexive(IDL.Vec(IDL.Nat));
+    testSub(IDL.Vec(IDL.Nat), IDL.Vec(IDL.Int));
+    testSubFail(IDL.Vec(IDL.Int), IDL.Vec(IDL.Nat));
+  });
+
+  describe('Subtyping on Options', () => {
+    // Because of some of the more "special" rules around Option types
```
crusso, line 988: special indeed

packages/candid/src/idl.test.ts
```diff
@@ -928,3 +933,191 @@ describe('IDL opt edge cases (embedded)', () => {
...
+
+    // Subtyping on individual fields
+    testSub(IDL.Record({ a: IDL.Nat }), IDL.Record({ a: IDL.Int }));
+    testSubFail(IDL.Record({ a: IDL.Int }), IDL.Record({ a: IDL.Nat }));
+
+    // Width subtyping
+    testSub(IDL.Record({ a: IDL.Nat, b: IDL.Nat }), IDL.Record({ a: IDL.Nat }));
+    testSubFail(IDL.Record({ a: IDL.Nat }), IDL.Record({ a: IDL.Nat, b: IDL.Nat }));
+
+    // Opt, Null, or Reserved fields are allowed to be missing
```
crusso, line 1009: Not just trailing ones, but in any position. Might be worth a couple more tests.

packages/candid/src/idl.ts
```diff
@@ -2032,3 +2069,170 @@ export function Func(args: Type[], ret: Type[], annotations: string[] = []): Fun
...
+
+/**
+ * Subtyping on Candid types t1 <: t2 (Exported for testing)
+ * @param t1 The potential subtype
+ * @param t2 The potential supertype
+ */
+export function subtype(t1: Type, t2: Type): boolean {
+  const relations = subtypeCache.copy();
+  const isSubtype = subtype_(relations, t1, t2);
+  if (isSubtype) {
```
crusso, line 2159: Don't you want to remember the failed tests too?

packages/candid/src/idl.ts
```diff
@@ -2032,3 +2069,170 @@ export function Func(args: Type[], ret: Type[], annotations: string[] = []): Fun
...
+    return new Relations(copy);
+  }
+
+  /// Returns whether we know for sure that a relation holds or doesn't (`true` or `false`), or
+  /// if we don't know yet (`undefined`)
+  known(t1: Type, t2: Type): boolean | undefined {
+    return this.rels.get(t1.name)?.get(t2.name);
+  }
+
+  addNegative(t1: Type, t2: Type) {
```
crusso, line 2096: This appears to be unused. Perhaps it should be use in subtype(t1,t2)?

crusso (commented):
Instead of using a global cache, how would you feel about passing the relation argument as the intial argument to decodeValue, along with the other bit of state, the `Pipe`.
Then maybe type could have a  binary decodeValue(b, t) that dispatches to the decodeValue(relates, b, t) with a fresh, local cache?

Also, constructing an object to enumberate the fields of variants and records seems a little inefficient. Since the underlying _fields arrays are sorted by label hash, can you just traverse the arrays directly, without the extra allocation?

packages/candid/src/idl.test.ts
```diff
@@ -928,3 +933,191 @@ describe('IDL opt edge cases (embedded)', () => {
...
+    const Even = IDL.Rec();
+    const Odd = IDL.Rec();
+    Even.fill(IDL.Tuple(Odd));
+    Odd.fill(IDL.Tuple(Even));
+
+    testSub(IDL.Tuple(Even), Odd);
+    testSub(IDL.Tuple(IDL.Tuple(Odd)), Odd);
+  });
+
+  describe('decoding function/service references', () => {
```
crusso, line 1084: Probably worth adding a test that a failing subtype test under an expected optional function/service type defaults to `null`.

crusso (commented):
Very nice and clean implementation!

packages/candid/src/idl.ts
```diff
@@ -1120,6 +1120,14 @@ export class RecordClass extends ConstructType<Record<string, any>> {
     return x;
   }
 
+  get fieldsAsObject(): Record<string, Type> {
```
crusso, line 1123: Convenient, but maybe inefficient too? Also, shouldn't we be comparing the fields by their hashes, not their names? If there's a collision, something might go subtly wrong.

packages/candid/src/idl.ts
```diff
@@ -1315,6 +1323,14 @@ export class VariantClass extends ConstructType<Record<string, any>> {
     }
     throw new Error('Variant has no data: ' + x);
   }
+
+  get alternativesAsObject(): Record<string, Type> {
```
crusso, line 1327: Same again. Convenient, but maybe inefficient too? Also, shouldn't we be comparing the fields by their hashes, not their names? If there's a collision, something might go subtly wrong.

packages/candid/src/idl.test.ts
christoph-dfinity: As these tests are checking records, there shouldn't be any notion of trailing?

I did add a bunch of test cases for the function check.

christoph-dfinity:
> Instead of using a global cache that needs to be reset carefully, how would you feel about passing the relation argument as the intial argument to decodeValue, along with the other bit of state, the Pipe.
Then maybe Type could have a binary decodeValue(b, t) that dispatches to the new ternary decodeValue(relations, b, t) with a fresh, local cache?

I got started on this by changing the signature of `Type.decodeValue` to include an optional argument for the cache (optional so I wouldn't cause a breaking change to the API), but found that after changing a few call sites, it was much easier to miss passing the cache in some place than to reset the global cache "carefully" in one place. Unfortunately you can't overload like you proposed in TypeScript.

> Also, constructing an object just to enumerate the fields of variants and records seems a little inefficient. Since the underlying _fields arrays (Array([string, Type]) appear to be sorted by hash of the string, can you just traverse the arrays directly, without the extra allocation? I guess that might mean recomputing he hashed before comparing the labels for relative order...

As discussed I decided to go for clarity over max performance in this case (matching the [Rust implementation](https://github.com/dfinity/candid/blob/30c388671462aecdc4a3a9753d50dc2e8208c200/rust/candid/src/types/subtype.rs#L81)). I think it's very unlikely to make a noticeable difference (especially with the subtyping cache around).

I think I addressed the rest of your feedback. PTAL

packages/candid/src/idl.ts
crusso: Perf can come later

packages/candid/src/idl.ts
```diff
@@ -1878,6 +1914,7 @@ export function decode(retTypes: Type[], bytes: ArrayBuffer): JsonValue[] {
     }
   });
 
+  resetSubtypeCache();
```
crusso, line 1917: Is it worth resetting the cache once done, to avoid leaking space?

christoph-dfinity: good call, done in c540f2522985aa685915f89abde2ac35254f6e8e

packages/candid/src/idl.ts
```diff
@@ -2032,3 +2069,174 @@ export function Func(args: Type[], ret: Type[], annotations: string[] = []): Fun
...
+    return this.rels.get(t1.name)?.get(t2.name);
+  }
+
+  addNegative(t1: Type, t2: Type) {
+    this.addNames(t1.name, t2.name, false);
+  }
+
+  add(t1: Type, t2: Type) {
+    this.addNames(t1.name, t2.name, true);
+  }
```
crusso, line 2108: Maybe we shouldn't worry but t.name is a property that is recomputed on each access.  It's quite expensive for record and variants. Do we want to address this, or leave it as an issue? It's a shame we can't use something else to key on.

christoph-dfinity:
I'll set up some benchmarks before trying to think too hard about what's fast or slow here... Thanks for the review I think we're good to merge for now.

christoph-dfinity:
@krpeacock Looks like I don't have the permissions to merge myself. Would you do me the honor?

PR #1004, opened 2025-04-17
refactor: use new error system and error codes

ilbertt:
# Description

Refactor the error system in the `@dfinity/agent` package to reflect Rust agent's implementation.
For now, I've added the error types and replaced all the internal `throw` statements with the new error system. This PR shouldn't break anything. In a follow-up PR, I'll refactor the agent and the actor errors in order to use the new error system across all the `@dfinity/agent` package.

# How Has This Been Tested?

Updated tests and assertions where already present.

# Checklist:

### DIFF
--- a/packages/agent/src/certificate.test.ts
+++ b/packages/agent/src/certificate.test.ts
@@ -16,6 +16,7 @@ import {
   CertificateNotAuthorizedError,
   CertificateTimeError,
   CertificateVerificationError,
+  ErrorKind,
 } from './errors';
 
 function label(str: string): ArrayBuffer {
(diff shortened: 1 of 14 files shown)


packages/agent/src/errors.ts
```diff
@@ -7,6 +7,400 @@ import {
...
+export class AgentErrorV2 extends Error {
+  public name = 'AgentError';
+
+  constructor(
+    public readonly code: ErrorCode,
+    public readonly kind: ErrorKindEnum,
+  ) {
+    super(
+      code.toErrorMessage(),
+      // @ts-expect-error - Error.cause is not supported in the Typescript version that we are using
```
ilbertt, line 47: Changing the TS config `lib` could break builds

packages/agent/src/certificate.test.ts
```diff
@@ -492,11 +515,25 @@ test('certificate verification fails on nested delegations', async () => {
...
       certificate: overlyNested,
       rootKey: fromHex(IC_ROOT_KEY),
       canisterId: canisterId,
-    }),
-  ).rejects.toThrow('Invalid certificate: Delegation certificates cannot be nested');
+    });
+  } catch (error) {
+    expect(error).toBeInstanceOf(ProtocolError);
+    expect(error.cause.code).toBeInstanceOf(CertificateHasTooManyDelegationsErrorCode);
+  }
```
ilbertt, line 538: I don't like this new way of asserting errors, but I couldn't find any other viable solution that didn't require creating custom Jest matchers

krpeacock (approved):
Idk if I'm just tired, but I didn't find any issues to take with the names or anything here

packages/agent/src/certificate.test.ts
krpeacock: yeah, I've done this before too. It's an okay pattern, and the alternative is to just thoroughly check the expected interface and prototype of the error

nathanosdev (approved):

packages/agent/src/errors.ts
ilbertt: I see we've changed the TS version in #1000, and I've anyway changed the AgentError in the next #1005, so we can resolve this

PR #1005, opened 2025-04-22
feat!: use new error system in Agent and Actor

ilbertt:
# Description

Use the error system introduced in #1004 in the Agent and the Actor classes. This enables us to remove some exported error classes in favor of the new error kinds and codes.

# How Has This Been Tested?

Updated assertions in tests to check for the specific error kind and code, where possible. Updated the `AgentError` tests.

# Checklist:


packages/agent/src/agent/http/index.ts
```diff
@@ -623,16 +632,21 @@ export class HttpAgent implements Agent {
...
-        error as HttpDetailsResponse,
-        toHex(requestId),
-        toHex(transformedRequest.body.sender_pubkey),
-        toHex(transformedRequest.body.sender_sig),
-        String(transformedRequest.body.content.ingress_expiry['_value']),
-      );
-      this.log.error(message, callError);
+      let callError: AgentError;
+      if (error instanceof AgentError) {
+        callError = UnknownError.fromCode(
```
ilbertt, line 637: Not sure if UnknownError is the proper error kind here, but I don't think we have a way to better determine the error kind here

packages/agent/src/agent/http/transforms.ts
```diff
@@ -50,6 +50,14 @@ export class Expiry {
...
     return lebEncode(this._value);
   }
+
+  public toString(): string {
+    return this._value.toString();
+  }
+
+  public toJSON(): string {
+    return this.toString();
+  }
```
ilbertt, line 60: I had to add this in order to serialize the `Expiry` class in the errors

ilbertt: @krpeacock would it be better to have it serialized as an object like we do with the Principal?

```suggestion

  public toString(): string {
    return this._value.toString();
  }

  public toJSON(): { _value: string } {
    return { _value: this.toString() };
  }
```

e2e/node/basic/mainnet.test.ts
```diff
@@ -201,7 +204,15 @@ test('it should allow you to set an incorrect root key', async () => {
...
 
-  expect(actor.whoami).rejects.toThrowError(`Invalid certificate:`);
+  expect.assertions(3);
+  try {
+    await actor.whoami();
+  } catch (error) {
+    expect(error).toBeInstanceOf(AgentError);
+    const errorCode = (error as AgentError).cause.code as WithRequestDetailsErrorCode;
+    expect(errorCode).toBeInstanceOf(WithRequestDetailsErrorCode);
+    expect(errorCode.caughtErrorCode).toBeInstanceOf(MissingSignatureErrorCode);
```
nathanosdev, line 214: I'm a bit conflicted about the extra layer of indirection... But maybe it depends on the usage. I'm not sure if we will expect devs to interact with the code or not.

Maybe we could remove the indirection by adding optional request details to `AgentError`.

packages/agent/src/errors.ts
```diff
@@ -322,41 +324,84 @@ export class TimeoutWaitingForResponseErrorCode extends ErrorCode {
 
   constructor(
     public readonly message: string,
-    public readonly requestId: RequestId,
-    public readonly status: RequestStatusResponseStatus,
+    public readonly requestId: RequestId | undefined = undefined,
+    public readonly status: RequestStatusResponseStatus | undefined = undefined,
```
nathanosdev, line 328: Small nit, but in (most) other places we don't explicitly assign `undefined` so I'd prefer to keep that consistent here

nathanosdev (approved):
For me this looks good. The errors make sense.

packages/agent/src/errors.ts
ilbertt: Fixed

packages/agent/src/agent/http/transforms.ts
krpeacock: yes, I think it should be an object for consistency. `toJSON` apis conventionally convert to a JSON-serializable format, instead of a serialized string (yes it's weird)

I'd add a matching `fromJSON` method for symmetry

packages/agent/src/errors.ts
```diff
@@ -20,47 +16,85 @@ enum ErrorKindEnum {
   Unknown = 'Unknown',
 }
 
-class ErrorCode {
-  public toErrorMessage(): string {
-    throw new Error('Not implemented');
-  }
+type RequestContext = {
```
krpeacock, line 19: let's export these types

krpeacock (approved):
two small changes requested, only nits. good to go once they're in

packages/agent/src/agent/http/transforms.ts
ilbertt: I'll make these changes in another PR

PR #1007, opened 2025-04-24
test: avoid skipping tests where possible, replace deprecated Jest assertions

ilbertt:
# Description

We should avoid skipping tests where possible. Where not possible, I've replaced some of them with either a `TODO` comment or a `.todo` Jest notation.

Additionally, I've replaced some of the deprecated Jest assertions.

# How Has This Been Tested?

CI

# Checklist:


tsconfig.json
```diff
@@ -8,7 +8,11 @@
     "rootDir": "."
   },
   "files": [],
-  "references": [{ "path": "packages/agent" }, { "path": "packages/identity" }],
+  "references": [
+    { "path": "packages/agent" },
+    { "path": "packages/identity" },
+    { "path": "packages/principal" }
```
ilbertt, line 14: Not sure if we should add the `principal` package reference here

krpeacock: It probably doesn't hurt? I remember the first two being necessary for some reason though

ilbertt: I've reconsidered and would rather not to touch the TS config now. Rolled back.

Issue #988, opened 2025-04-05
Failed to decode CBOR (Failed to parse, Undeterminated nesting)


pu0238:
**Describe the bug**  
I am using "@dfinity/agent": "^2.4.0" and sometimes, while doing communication with the IC mainnet, I receive errors like: *Failed to parse*, *Undetermined nesting* — even though the whole transaction was fine. I rely heavily on those responses, so it's messing up the flow of the app and may cause serious issues in my application.

After sending a transaction and once it is computed, I receive an error, even though the full response can be successfully decoded using any CBOR tool (e.g., [https://cbor.me/](https://cbor.me/)):

[...]

PR #1015, opened 2025-05-12
feat: use new cbor library

nathanosdev:
# Description

Use the `@dfinity/cbor` package instead of the `borc` and `simple-cbor` dependencies. This reduces a lot the bundle size.

`BigInt`s are now supported natively by the new cbor package, closes #624  

# How Has This Been Tested?

Same tests should still pass, plus a few more tests have been added in the `cbor.test.ts` file.

# Checklist:


nathanosdev:
The pipelines will continue to fail until the CBOR package is published and the `package.json` is updated.

ilbertt:
@nathanosdev, does this fix #988?

packages/agent/src/cbor.ts
```diff
@@ -1,92 +1,10 @@
...
-const serializer = SelfDescribeCborSerializer.withDefaultEncoders(true);
-serializer.addEncoder(new PrincipalEncoder());
-serializer.addEncoder(new BufferEncoder());
-serializer.addEncoder(new BigIntEncoder());
-
-export enum CborTag {
-  Uint64LittleEndian = 71,
-  Semantic = 55799,
+function hasCborReplacerMethod(value: unknown): value is { toCborReplacer: () => cbor.CborValue } {
+  return typeof value === 'object' && value !== null && 'toCborReplacer' in value;
```
ilbertt, line 7: We cannot import elements from other packages (e.g. `Delegation` from `@dfinity/identity`), so we need to somehow "agree" on the other packages to expose this `toCborReplacer` method

packages/agent/src/cbor.ts
```diff
@@ -1,126 +1,46 @@
...
-  for (let i = 0; i < len; i++) {
-    res = res * BigInt(0x100) + BigInt(buf[i]);
-  }
-
-  return res;
-}
+export function encode(value: unknown): Uint8Array {
+  try {
+    return cbor.encodeWithSelfDescribedTag(value, value => {
+      if (value instanceof Principal) {
```
krpeacock, line 28: Here I think we should have a more sophisticated check - the principal may have been passed from a different version of the library, so we can lean on the `_isPrincipal` tag. Same goes for Expiry below

krpeacock: looks good

krpeacock (approved):

krpeacock (approved):

PR #1022, opened 2025-05-16
refactor: use hex utils from `@noble/hashes`

ilbertt:
# Description

Minor changes

# How Has This Been Tested?

Same tests should pass.

# Checklist:

### DIFF
--- a/packages/identity-secp256k1/src/secp256k1.test.ts
+++ b/packages/identity-secp256k1/src/secp256k1.test.ts
@@ -51,26 +51,11 @@ describe('Secp256k1PublicKey Tests', () => {
     });
   });
 
-  const hexRe = new RegExp(/^[0-9a-fA-F]+$/);
-  function fromHex(hex: string): Uint8Array {
-    if (!hexRe.test(hex)) {
-      throw new Error('Invalid hexadecimal string.');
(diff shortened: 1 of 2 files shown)


krpeacock (approved):

peterpeterparker:
> @dfinity/principal 	4.39 KB (+4.96% 🔺)

The agent is already too big - wouldn't it make sense not to make it even bigger unless there is a hurge to replace those utils?

ilbertt:
> The agent is already too big - wouldn't it make sense not to make it even bigger unless there is a hurge to replace those utils?

We are thinking to reorganize the dependencies/shared code anyway. This PR was more to clean the code, I don't think that's the definitive result

peterpeterparker:
Will this change be released before the actual reorganization takes place? If so, it means a larger bundle will be shipped in the meantime. Personally, I don’t think any size increase is a good idea - even if it’s just a 5% bump on the principal - considering that the agent is already way too heavy and dapps have to pay the cost. That said, if there’s no intermediate release and the final result leads to a smaller bundle, sure that sounds totally fine.

Anyway, not my call. I just wanted to flag the size increase in case the team hadn't noticed it.

PR #931, opened 2024-09-16
fix: remove the unnecessary 'await import' (#930)

zensh:
# Description

`await import` is incompatible with the ServiceWorker script. This can cause the error: "Uncaught SyntaxError: Cannot use 'import.meta' outside a module."

Fixes # (issue)
https://github.com/dfinity/agent-js/issues/930

# Checklist:

- [ ] My changes follow the guidelines in [CONTRIBUTING.md](https://github.com/dfinity/agent-js/blob/main/CONTRIBUTING.md).
- [ ] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #1026, opened 2025-05-23
refactor: update typescript config

ilbertt:
# Description

Update TypeScript config to a more modern approach.

Additionally, it removes the `demos/sample-javascript` folder, which seems to be outdated.

Closes #931.

# How Has This Been Tested?

Same tests should still pass.

# Checklist:


krpeacock:
I don't actually care that much about the jest files - it's annoying that there's any trouble with them specifically though

ilbertt (commented):
I've skipped the refactoring of `packages/bls-verify`, is it's deprecated

stats.json
ilbertt, line 1: @krpeacock do we still need this file?

ilbertt:
Although a lot of files have changed, most of them are config files (either adapted to the new TS config or removed as not needed anymore). The source files have been changed to adapt to the new TS config: I've added `type` notation on imports, removed unnecessary comments and prefixed unused variables with `_`.

ilbertt:
I've tried importing the agent locally in a Node v20.18 (pre-ESM support), Node v20.19 (post-ESM support, see https://nodejs.org/en/blog/release/v20.19.0/#requireesm-is-now-enabled-by-default) and Node v22 with both `import` and `require` notations, and it seems to work properly.

packages/agent/src/agent/http/http.test.ts
```diff
@@ -702,7 +702,7 @@ test('should adjust the Expiry if the clock is more than 30 seconds behind', asy
 
   const requestBody: any = cbor.decode(mockFetch.mock.calls[0][1].body);
 
-  expect(requestBody.content.ingress_expiry).toMatchInlineSnapshot(`1260000000000`);
+  expect(requestBody.content.ingress_expiry).toMatchInlineSnapshot(`1200000000000`);
```
nathanosdev: Why did this snapshot change?>

ilbertt: Oh, good that you caught it! I'll give it a look again, still not sure what's happening there

ilbertt: @nathanosdev it should be fixed now, see https://github.com/dfinity/agent-js/pull/1026/commits/be9a24dd92d105320e3eec071bb8b1ab6ec14b2b

tsconfig.json
```diff
@@ -1,15 +1,41 @@
...
-    // Required for buffer-pipe
+    "rootDir": ".",
+    "allowJs": false,
+    "declaration": true,
+    "forceConsistentCasingInFileNames": true,
+    "incremental": true,
+    "resolveJsonModule": true,
+    "sourceMap": true,
+    "inlineSources": true,
+    "useDefineForClassFields": false,
```
krpeacock, line 11: Does this flag need to exist in perpetuity, or only for as long as we target ES2022?

packages/assets/src/index.ts
```diff
@@ -394,8 +392,6 @@ class Asset {
   constructor(
     private readonly _actor: ActorSubclass<AssetsCanisterRecord>,
     private readonly _limit: LimitFn,
-    private readonly _maxSingleFileSize: number,
-    private readonly _maxChunkSize: number,
```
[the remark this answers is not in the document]
krpeacock: Can confirm

e2e/node/basic/watermark.test.ts
```diff
@@ -8,7 +8,7 @@ class FetchProxy {
   #calls = 0;
   #replyIndex: number | null = null;
 
-  async fetch(...args): Promise<Response> {
+  async fetch(...args: Parameters<typeof global.fetch>): Promise<Response> {
```
krpeacock, line 11: I wasn't aware of this feature! It's convenient

krpeacock (approved):
Thanks for picking up such a hefty refactor!

tsconfig.json
ilbertt: We could probably set it to `true` already, but it was causing some issues, as you can also see from https://github.com/microsoft/TypeScript/issues/48814

PR #1029, opened 2025-06-04
ci: test docs builds

ilbertt:
# Description

Test that the docs build properly on every pull request.

# How Has This Been Tested?

CI passes

# Checklist:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,9 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
       - uses: ./.github/actions/prepare
         with:
           node-version: 22
+
+      # - run: npm run build # TODO: uncomment this
+
       - name: Generate docs
         run: npm run make:docs
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
 
       - uses: ./.github/actions/prepare
         with:
-          node-version: 22
+          node-version-file: '.nvmrc'
 
       - run: npm run build
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,30 @@
+name: Test Docs Build
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+
+jobs:
+  test:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,34 @@
+name: Test Docs Build
+
+on:
+  pull_request:
+    types:
```
nathanosdev, line 5: If we exclude `types` then it'll just run on all PRs. I know it's consistent with the other actions for now though so it's fine.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/prepare
+        with:
+          node-version: 22
```
nathanosdev: Why not use the `node-version-file` file here? Maybe the `prepare` action could just default to using `.nvmrc`.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - run: npm run build
```
nathanosdev, line 21: Do we really need to build before building the docs? Are the docs are generated from the TypeScript code?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Generate docs
+        run: npm run make:docs
+
+  aggregate:
```
nathanosdev, line 26: Without this, I guess it's not failing the pipeline when the docs build fails?

.github/workflows/docs.yml
ilbertt: Yeah, I just kept it for consistency. This way, when we change it only once

.github/workflows/docs.yml
ilbertt: Yes, if we don't build it, it fails. See https://github.com/dfinity/agent-js/actions/runs/15441423340/job/43459923540

.github/workflows/docs.yml
ilbertt: I think it fails anyway, but again I kept it for consistency with the other actions

.github/workflows/docs.yml
ilbertt: Fixed

PR #1034, opened 2025-06-18
fix: mark `@noble/hashes` as a dependency rather than a dev dependency

hpeebles:
# Description

I have just updated OpenChat to use v3.0.0-beta.0, but I had to manually add a dependency on `@noble/hashes` because prior to this PR it is only listed as a dev dependency of `@dfinity/identity` even though it is used at run time.

It is not a dev dependency of any other packages in this workspace.

# How Has This Been Tested?

It only moves a dependency from `devDependencies` to `dependencies`, so testing is not required.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15741,10 +15741,10 @@
       "version": "2.4.1",
       "license": "Apache-2.0",
       "dependencies": {
-        "@noble/curves": "^1.2.0"
+        "@noble/curves": "^1.2.0",
+        "@noble/hashes": "^1.8.0"
       },
       "devDependencies": {
-        "@noble/hashes": "^1.8.0",
         "@peculiar/webcrypto": "^1.4.0",
         "@typescript-eslint/eslint-plugin": "^5.30.5",
         "@typescript-eslint/parser": "^5.30.5",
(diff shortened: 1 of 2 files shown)


ilbertt:
@hpeebles good catch. Can you run `npm install` again to update the `package-lock.json` as well?

ilbertt:
> I can make the changelog update

I'm fine even without a changelog update in this case, as this is implicitly included in this changelog point:
> - feat!: replaces fromHex, toHex, concat utils with bytesToHex, hexToBytes, and concatBytes from @noble/hashes/utils respectively, to take advantage of existing dependencies.

ilbertt:
> I wonder if it should be peer dependencies?
> 
> * Are they used in any other agent related libs?
> * They might be use by the consumer as well? e.g. `@dfinity/ckbtc`, `@dfinity/ledger-icp` or `@dfinity/nns`
> 
> Just thinking out at loud.

I'd prefer to keep them as dependencies to avoid having to install a lot of libraries just to make a basic app work. We are anyway planning to refactor the packages structure, and that will solve these kinds of issues anyway.

peterpeterparker:
> I'd prefer to keep them as dependencies to avoid having to install a lot of libraries just to make a basic app work.

npm automatically install peer dependencies.

PR #1035, opened 2025-06-18
chore: update `@noble/*` dependencies

ilbertt:
# Description

Update the `@noble/curves` and `@noble/hashes` dependencies. Following the suggestion in https://github.com/dfinity/agent-js/pull/1034#issuecomment-2983647447, sets them as peer dependencies where needed.

# How Has This Been Tested?

Same tests should still pass.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,17 +117,6 @@
         "node": ">=10.10.0"
       }
     },
-    "e2e/node/node_modules/@noble/hashes": {
-      "version": "1.8.0",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": "^14.21.3 || >=16"
-      },
-      "funding": {
-        "url": "https://paulmillr.com/funding/"
-      }
-    },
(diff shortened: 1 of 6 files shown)


peterpeterparker:
Coolio.

First of all, note that my previous comment about peer dependencies was really more of me "thinking out loud" rather than suggesting an immediate call to action. Just saying in case you interpreted it as something critical, which wasn’t my intention.

That said, since all the bundles are growing, I wonder if we shouldn't exclude the related peer dependencies from the `size-limit` configuration for those packages?

nathanosdev (approved):

ilbertt:
> That said, since all the bundles are growing, I wonder if we shouldn't exclude the related peer dependencies from the `size-limit` configuration for those packages?

You're right, but the upcoming packages refactor will also change how we bundle them. So I would not worry about it now

peterpeterparker:
> So I would not worry about it now

Capture d’écran 2025-06-18 à 15 12 14

PR #1037, opened 2025-06-24
chore: fix incremental typescript config

nathanosdev:
# Description

The Typescript build cache files are better stored in the `lib` directory in my opinion. If the `lib` directory is deleted, you should also delete the `.tsbuildinfo` file. So I opted to move the file back to it's default location inside of `outDir` (from `tsconfig.json`) and then ignore it for publishing with `.npmignore`.

We should keep the references in `tsconfig.json` in sync with any peer dependencies from `package.json` that exist in the current workspace.

This exposed some issues in the `package.json` exports that I've also fixed, and I also just aligned some inconsistencies where possible.

The IDE support is better now. CTRL/CMD + click on an element from a peer package will go to the correct line on the original `.ts` file in the `src` now instead of the transpiled `.d.ts` file in the `lib` folder.

# How Has This Been Tested?

This only affects build related files, so the CI pipelines are enough.

# Checklist:


packages/bls-verify/.gitignore
nathanosdev, line 1: All of the deletions related to `bls-verify` were just not used.

nathanosdev:
This changes doesn't make any difference to the CI pipeline running time, but it should provide noticeable improvements in performance when developing locally through the terminal and the IDE.

packages/agent/tsconfig.cjs.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": ["./tsconfig.json", "../../tsconfig.cjs.json"],
   "compilerOptions": {
-    "baseUrl": ".",
+    "baseUrl": "./src",
+    "rootDir": "./src",
     "outDir": "./lib/cjs",
-    "rootDir": "./src"
-  }
+    "lib": ["DOM"]
```
nathanosdev, line 7: I wasn't sure actually why I had to add this specifically to the CJS config after this change...

packages/auth-client/package.json
```diff
@@ -32,15 +32,13 @@
...
-    }
-  },
+  "main": "./lib/cjs/index.js",
+  "module": "./lib/esm/index.js",
+  "browser": "./lib/esm/index.js",
+  "unpkg": "./lib/esm/index.js",
+  "types": "./lib/esm/index.d.ts",
   "scripts": {
-    "build": "tsc -b",
+    "build": "tsc -b && tsc -p tsconfig.cjs.json",
```
nathanosdev: I'm happy to revert this if we specifically don't want to release CJS here.

ilbertt: The idea is that, since this package is only used in browser environments, we don't need to support CommonJS

packages/agent/tsconfig.json
```diff
@@ -2,20 +2,11 @@
...
-    "baseUrl": "./",
-    "outDir": "lib/esm",
+    "baseUrl": "./src",
     "rootDir": "./src",
-    "tsBuildInfoFile": "./build_info.json"
+    "outDir": "./lib/esm"
   },
-  "include": ["types/*", "src/**/*.ts"],
-  "exclude": [
-    "**/src/**/__certificates__/**",
```
ilbertt, line 12: Removing`__certificates__` from the `exclude` config is going to add the `packages/agent/src/agent/http/__certificates__/goldenCertificates.ts`, isn't it?

packages/bls-verify/.gitignore
ilbertt: Do we want to keep touching the `bls-verify` package? I haven't touched it in #1026 as well. I'd rather remove every file inside this folder except the readme that says that the package is deprecated.

packages/agent/tsconfig.cjs.json
ilbertt: Weird... Can't we just set it in the root `tsconfig.cjs.json` file then? So that we avoid repeating it in all the packages configs

packages/auth-client/tsconfig.cjs.json
ilbertt, line 1: I wouldn't add CommonJS config for this package, see also https://github.com/dfinity/agent-js/pull/1037/files#r2163915547

ilbertt (changes requested):

packages/bls-verify/.gitignore
nathanosdev: That's fair, I'll do that in a follow up PR.

packages/agent/tsconfig.cjs.json
nathanosdev: It's not needed by every package so I'd leave it as an explicit choice on a per-package basis to support that. Otherwise we will have types for DOM APIs available in all libs, even the ones only meant for NodeJS.

packages/auth-client/package.json
nathanosdev: Thanks, I'll remove.

packages/agent/tsconfig.json
nathanosdev: Good catch, thanks.

packages/auth-client/package.json
```diff
@@ -32,13 +32,11 @@
...
     "api"
   ],
-  "unpkg": "./lib/index",
-  "exports": {
-    ".": {
-      "import": "./lib/index.js",
-      "require": "./lib/index.js"
-    }
-  },
+  "main": "./lib/cjs/index.js",
```
ilbertt: There's a typo here, we don't have cjs in this package

ilbertt (approved):

packages/agent/tsconfig.cjs.json
ilbertt: @nathanosdev I don't think we have any package that is only meant for NodeJS at the moment, and probably not in the near future. But I'm fine with leaving as it is now as well.

packages/auth-client/package.json
nathanosdev: Thanks!

PR #1038, opened 2025-06-24
chore: switch to pnpm

nathanosdev:
# Description

This PR migrates from npm to pnpm. Any common dev dependencies have been consolidated in the root package.json. dependencies and peerDependencies are still managed in the individual projects.

There's also a migration to a newer version of eslint. Some new files are being linted now that were not being linted before, they have some issues fixed.

Some unused projects are also removed in this PR.

# How Has This Been Tested?

I pressed the delete key, then I checked with my eyes if the files were really gone.

# Checklist:


ilbertt (approved):
Nice! How about changing the `packages/bls-verify/README.md` file like you did in https://github.com/dfinity/agent-js/pull/1037/files#diff-871487222e076ccbebcb01206cffe37dea88ae2e3003b112cf8ba12ae42d2609:

```markdown
# @dfinity/bls-verify

`**Warning** this package is deprecated`
```

.github/workflows/e2e-tests.yml
```diff
@@ -25,44 +25,22 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/prepare
-        with:
-          node-version: ${{ matrix.node }}
 
-      - name: Cache node modules
```
nathanosdev, line 32: Caching of node modules is already handled by `setup-pnpm`

.gitignore
```diff
@@ -127,7 +127,6 @@ docs/identity-secp256k1
 docs/identity-ledgerhq
 docs/principal
 docs/candid
-docs/bls-verify
```
nathanosdev, line 130: Everything except the readme of this project was removed.

bin/version.ts
nathanosdev, line 1: Lint fixes.

e2e/node/basic/mainnet.test.ts
```diff
@@ -214,28 +212,3 @@ test('it should allow you to set an incorrect root key', async () => {
     expect(errorCode.requestContext).toBeDefined();
   }
 });
-
-test('should allow you to sync time when the system time is over 5 minutes apart from the replica time', async () => {
```
nathanosdev, line 218: This test broke somehow, but anyway we have other tests in the `syncTime.ts` file that cover this scenario using spies on the calls to replica, rather than spies on logs.

e2e/node/dfx.json
```diff
@@ -2,7 +2,8 @@
   "canisters": {
     "assets": {
       "type": "assets",
-      "source": ["canisters/assets"]
+      "source": ["canisters/assets"],
+      "build": ""
```
nathanosdev, line 6: The default command for the asset canister is `npm run build`, which fails now. We don't need to build anything for this.

e2e/node/package.json
```diff
@@ -1,48 +1,26 @@
...
+    "@dfinity/identity-secp256k1": "workspace:*",
+    "@dfinity/principal": "workspace:*"
   },
   "devDependencies": {
+    "@noble/curves": "^1.9.2",
     "@noble/hashes": "^1.8.0",
-    "@peculiar/webcrypto": "^1.4.0",
     "@tsconfig/node22": "^22.0.2",
     "@types/express": "^5.0.2",
-    "@types/isomorphic-fetch": "^0.0.36",
```
nathanosdev, line 29: Dependencies have been consolidated into the root, de-duped or just removed entirely if they're not used.

e2e/node/test-setup.ts
```diff
@@ -9,13 +8,11 @@
 
 // Node.js in a similar way.
 import { TextEncoder, TextDecoder } from 'text-encoding';
-import fetch from 'isomorphic-fetch';
```
nathanosdev, line 12: Who needs fetch these days?

eslint.config.mjs
nathanosdev, line 1: This does the same thing as before, but it covers a few extra files now. Such as the `bin` directory and test files in each project.

packages/bls-verify/.eslintignore
nathanosdev, line 1: All unused, only the readme remains.

packages/use-auth-client/examples/auth-demo/.gitignore
nathanosdev, line 1: Deleted completely, not used.

README.md
```diff
@@ -52,19 +50,17 @@ We expect you to have the replica running on port `4943`. If you are using a dif
...
 
-We use `npm` to manage this repo and its packages. A few useful
+We use `pnpm` to manage this repo and its packages. A few useful
 commands to keep in mind;
 
-- To run the unit tests locally, you can use `npm run test`.
-- To run e2e tests, you can use `npm run e2e`. **WARNING:** You need to have a running
+- To run the unit tests locally, you can use `pnpm test`.
+- To run e2e tests, you can use `pnpm e2e`. **WARNING:** You need to have a running
   replica locally. In our CI runs, we use the `ic-ref` which is not (at this time) available
```
ilbertt: I think the text starting from _In our CI runs_ can be removed.

README.md
```diff
@@ -77,12 +73,12 @@ The following scripts can be found in [./bin](./bin):
 
 - update-management-idl - Update the management canister IDL in @dfinity/agent
 
-Monorepo-related scripts run in this order, but are usually invoked by `npm install`:
+Monorepo-related scripts run in this order, but are usually invoked by `pnpm i`:
 
-- npm-postinstall - Run with `npm run postinstall` in this monorepo package.
+- npm-postinstall - Run with `pnpm postinstall` in this monorepo package.
```
ilbertt: Maybe call it like this instead?
```suggestion
- postinstall - Run with `pnpm postinstall` in this monorepo package.
```

e2e/node/package.json
ilbertt: One thing that I don't like with this approach is that if you want to extract and isolate a package, most of the dev deps would be missing and hence the package would be broken (e.g you can't properly build or test it without going to find all the dev deps missing). But since there's no way to specify that the package manager should use deps from the root (like you do in Rust/Cargo workspaces), I'm fine with this approach

package.json
```diff
@@ -3,74 +3,65 @@
...
-    "webcrypto-core": "^1.7.7",
-    "yargs": "^17.7.2",
-    "zx": "^8.1.2"
+    "vitest": "^3.2.4"
   },
+  "packageManager": "pnpm@10.10.0+sha512.d615db246fe70f25dcfea6d8d73dee782ce23e2245e3c4f6f888249fb568149318637dca73c2c5c8ef2a4ca0d5657fb9567188bfab47f566d1ee6ce987815c39",
   "engines": {
     "node": "^20 || ^22",
-    "npm": ">=8"
+    "pnpm": ">=10"
```
ilbertt: Maybe add a notice as usual?
```suggestion
    "pnpm": ">=10",
    "npm": "please use pnpm"
```

package.json
```diff
@@ -87,37 +78,42 @@
     {
       "name": "@dfinity/agent",
       "path": "./packages/agent/dist/index.js",
-      "limit": "105 kB"
-    },
-    {
-      "name": "@dfinity/candid",
-      "path": "./packages/candid/dist/index.js",
-      "limit": "20 kb"
+      "limit": "50 kB"
```
ilbertt, line 82: Nice that you've updated the size limits

packages/assets/tsconfig.json
```diff
@@ -8,5 +8,5 @@
   },
   "include": ["./src/**/*.ts"],
   "exclude": ["./src/**/*.test.ts"],
-  "references": [{ "path": "../agent" }, { "path": "../principal" }]
+  "references": [{ "path": "../agent" }, { "path": "../principal" }, { "path": "../candid" }]
```
ilbertt, line 11: I find it annoying that you have to specify all the deps in each package both in the `package.json` and here, but there's no other way probably

pnpm-workspace.yaml
```diff
@@ -0,0 +1,5 @@
+packages:
+  - e2e/*
+  - packages/*
```
ilbertt: I'd prefer to be explicit here, so that we are enforced to specify new packages that we really want to include here and keep track when and if we move existing packages. But I'm fine with this approach as well.

e2e/node/package.json
nathanosdev: That's a good point.

Maybe we could discuss using this: https://pnpm.io/catalogs. I don't love the idea at first because it's non-standard, but it helps solve the issue you're describing.

packages/assets/tsconfig.json
nathanosdev: I agree, it's not ideal. We have `pnpm` and `tsc` both doing workspace level stuff, but independently from each other. The alternative would be to use a more invasive monorepo management tool that handles everything.

pnpm-workspace.yaml
nathanosdev: I'm usually explicit in cases where only some sub directories are JS libraries (others may be Rust). It felt unnecessary in this case, but I'm also fine in either direction so I can change it.

e2e/node/package.json
ilbertt: https://pnpm.io/catalogs sounds powerful, but I also don't like the idea of non-standard package manager stuff. Let's leave it as you did for now.

packages/assets/tsconfig.json
ilbertt: Let's stick to `pnpm` and `tsc` for now. This is going to improve once we have a monopackage.

ilbertt (approved):

PR #1041, opened 2025-06-26
feat: add @dfinity/icp monopackage

nathanosdev:
# Description

This PR adds the `@dfinity/icp` monopackage, which re-exports every other package in this workspace.

# Checklist:


README.md
```diff
@@ -50,8 +50,8 @@ We expect you to have the replica running on port `4943`. If you are using a dif
 To run the e2e browser tests, you can run
 
 ```bash
-pnpm -F @e2e/browser setup
-pnpm -F @e2e/browser e2e
+pnpm run -F @e2e/browser setup
+pnpm run -F @e2e/browser e2e
 ```
```
ilbertt, line 55: How about we remove the e2e browser section (lines 48-55)?

packages/icp/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "composite": true,
+    "baseUrl": "./src",
+    "rootDir": "./src",
+    "outDir": "./lib/esm"
+  },
+  "include": ["./src/**/*.ts"],
+  "exclude": ["./src/**/*.test.ts"],
+  "references": [{ "path": "../candid" }, { "path": "../principal" }]
```
ilbertt: Why aren't all the packages referenced here?

packages/icp/package.json
```diff
@@ -0,0 +1,98 @@
+{
+  "name": "@dfinity/icp",
+  "version": "3.0.0-beta.1",
```
ilbertt: Shouldn't this package have a separate versioning?

packages/icp/README.md
ilbertt, line 1: Do you plan to add a readme in a follow-up PR?

packages/icp/src/use-auth-client/index.ts
ilbertt, line 1: I'd rather keep this as a separate package, as this is a React-specific package. Otherwise, we can think of introducing a `react` submodule under which to put the `use-auth-client` module.

ilbertt (changes requested):

README.md
nathanosdev: Sure, I'll follow up with this.

packages/icp/tsconfig.json
nathanosdev: Good catch, it should be just `icp` and `agent`.

packages/icp/package.json
nathanosdev: I think the versioning should continue as-is, because it's still the same code. Whether the code stays outside the `icp` folder or moves inside, it's still the same code so I think starting a new version history here would make it more confusing.

packages/icp/README.md
nathanosdev: Yep. More docs changes in general too.

packages/icp/src/use-auth-client/index.ts
nathanosdev: Makes sense. I'll just remove it for now. If more react libs come then maybe a react submodule could make sense.

e2e/node/basic/canisterStatus.test.ts
```diff
@@ -1,5 +1,5 @@
-import { AgentError, CanisterStatus, HttpAgent } from '@dfinity/agent';
-import { Principal } from '@dfinity/principal';
+import { AgentError, CanisterStatus, HttpAgent } from '@dfinity/icp/agent';
```
krpeacock, line 1: see, I'm already thinking ahead to how I would rather have something like `CanisterStatus` live at `@dfinity/icp/canister_status`

e2e/node/package.json
```diff
@@ -3,18 +3,15 @@
   "private": true,
   "type": "module",
   "scripts": {
+    "build": "tsc -b",
```
krpeacock, line 6: Do the e2e tests need to be built before they can be tested? Or is this a pnpm issue?

krpeacock, line 6: Do the e2e tests need to be built before they can be tested? Or is this a pnpm issue?

e2e/node/package.json
```diff
@@ -3,18 +3,15 @@
...
   },
   "dependencies": {
-    "@dfinity/agent": "workspace:*",
-    "@dfinity/assets": "workspace:*",
-    "@dfinity/candid": "workspace:*",
-    "@dfinity/identity": "workspace:*",
-    "@dfinity/identity-secp256k1": "workspace:*",
-    "@dfinity/principal": "workspace:*"
+    "@dfinity/icp": "workspace:*",
+    "@dfinity/agent": "workspace:*"
```
krpeacock, line 14: Why is agent still needed specifically?

packages/icp/README.md
krpeacock, line 1: This probably shouldn't be blank

krpeacock (approved):

packages/icp/package.json
ilbertt: @nathanosdev I don't agree, the `@dfinity/icp` package is meant to be a whole separate thing, and that's why we should start a new versioning specifically. As a consequence, what we do inside to produce and expose the code shouldn't matter.

e2e/node/package.json
nathanosdev: This won't build the e2e tests, just type check them (the tsconfig is set to not emit). It will build the libraries that the e2e tests depend on, if they are not already built.

e2e/node/package.json
nathanosdev: This is only the case with projects that use `pnpm`. `pnpm` doesn't allow your code to access peer dependencies, where `npm` would allow. This stricter setup requires you to explicitly list every dependency that your application code will access.

In this case it's the `didc` output that's still using `@dfinity/agent`. Since the new `didc` generation target will use `@dfinity/icp` we can remove this from the `package.json` once we have the new DFX release and we update the agent repo to use it.

packages/icp/README.md
nathanosdev: It definitely shouldn't. I intend to follow up with more on docs and pipelines in more PRs to keep the scope of this one smaller.

packages/icp/package.json
nathanosdev: Updated to `1.0.0-beta.0`.

ilbertt (approved):

PR #1042, opened 2025-06-30
ci: cancel CI tests that become obsolete after a new commit is pushed in an open PR

kitsiosk:
# Description

### Change Summary
Currently, if a PR is open and a push happens, the unit-tests workflow will start running. If, shortly after a subsequent push on the *same* PR happens, the unit-tests workflow will start running again without cancelling the previous (now obsolete) run. With these changes, the first run will be cancelled, thus saving compute resources (see below for quantity) without sacrificing functionality, since the second run will contain the changes from the first push as well.

### Context
Hi,

We are a team of [researchers](https://www.ifi.uzh.ch/en/zest.html) from University of Zurich and we are currently working on energy optimizations in Github Actions workflows.

Here is an example of the behaviour described above concerning PR #1000: the commit `e2788bc2dfc9f700e6de4ad2b2086181b2f56bde` triggered [this](https://github.com/dfinity/agent-js/actions/runs/14539513883) workflow run, and one minute later the commit `3f3008c11b2a5eb0f2814c2dfba4783fec32fb9d`, that happened on top of the first commit, triggered [this](https://github.com/dfinity/agent-js/actions/runs/14539534936) workflow. Both workflows ran till the end, spending ~6 CPU minutes each. With the proposed changes, the first run would be cancelled, hence saving ~5 CPU minutes and clearing the queue for other workflows. Note that this is an example of a single concurrent run, and the accumulated gain for all PRs would be higher.

Kindly let us know (here or in the email below) if you would like more details on our offline evaluation, if you want to reject the proposed changes for other reasons, or if you have any question whatsoever.

Best regards,
[Konstantinos Kitsios](https://www.ifi.uzh.ch/en/zest/team/konstantinos_kitsios.html)
konstantinos.kitsios@uzh.ch

# How Has This Been Tested?

No test was run, since the three lines added come directly from the from the official [documentation](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow).

# Checklist:


krpeacock (dismissed):
makes sense to me

.github/workflows/mitm.yml
```diff
@@ -9,6 +9,10 @@ on:
       - synchronize
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.event_name }}-${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }} # this will cancel only PRs
```
nathanosdev, line 14: We can simplify here by removing the `workflow_dispatch` trigger (it's currently not used) and changing `cancel-in-progress` to `true`.

nathanosdev (dismissed):
This is a nice improvement, thanks!

nathanosdev (approved):

PR #1043, opened 2025-07-01
chore: updates docs and package config

nathanosdev:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

# Checklist:

- [ ] My changes follow the guidelines in [CONTRIBUTING.md](https://github.com/dfinity/agent-js/blob/main/CONTRIBUTING.md).
- [ ] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #1045, opened 2025-07-01
feat: remove watermark checks

### LINKED ISSUE #980: [Feature Request] Allow disabling watermark checking
Timestamp failed to pass the watermark after retrying the configured 3 times
see https://forum.dfinity.org/t/timestamp-failed-to-pass-the-watermark-after-retrying-the-configured-3-times/29180/21?u=let4be

I want low latency in my calls and I don't care about getting slightly stale data

**Describe alternatives you've considered**
We could try increase max retries to eventually get the response, but this would incur high latence which is bad for my use case

**Additional context**
Image

ilbertt:
# Description

Removes the watermark checks in favor of checking the query signature's timestamp against the ingress expiry config parameter.

Closes #980

# How Has This Been Tested?

Added tests to `e2e/node` package.

# Checklist:


packages/agent/src/agent/http/types.ts
ilbertt, line 1: Also fixed in https://github.com/dfinity/agent-js/pull/1043

packages/agent/src/agent/api.ts
ilbertt, line 1: Fixed also in https://github.com/dfinity/agent-js/pull/1043

packages/agent/src/agent/http/index.ts
```diff
@@ -909,24 +874,24 @@ export class HttpAgent implements Agent {
...
-    let queryResult;
     const id = await (identity ?? this.#identity);
     if (!id) {
       throw ExternalError.fromCode(new IdentityInvalidErrorCode());
     }
 
     const canister = Principal.from(canisterId);
     const sender = id.getPrincipal();
+    // No need to adjust the time here because the query response comes from a single node,
+    // and hence it can be different from the synced network time.
```
mraszyk: I didn't get this comment.

packages/agent/src/agent/http/index.ts
```diff
@@ -678,13 +668,8 @@ export class HttpAgent implements Agent {
...
     if (delay === null) {
-      throw ProtocolError.fromCode(
-        new TimeoutWaitingForResponseErrorCode(
-          `Timestamp failed to pass the watermark after retrying the configured ${
-            this.#retryTimes
-          } times. We cannot guarantee the integrity of the response since it could be a replay attack.`,
-          requestId,
-        ),
+      throw TrustError.fromCode(
+        new CertificateOutdatedErrorCode(this.#maxIngressExpiryInMinutes, requestId, tries),
```
mraszyk: Can't we get here also due other error conditions, e.g., networking issues?

packages/agent/src/agent/http/index.ts
```diff
@@ -747,51 +732,35 @@ export class HttpAgent implements Agent {
...
-      this.log.error('Timestamp is below', error, {
-        timestamp,
-        waterMark: this.waterMark,
+    // If the timestamp is older than the max ingress expiry, retry the request up to the retry limit.
+    // We don't have to consider the #timeDiffMsecs here because the response comes from a single node,
+    // and hence it can be different from the network time.
+    if (Date.now() - timestampInMs > this.#maxIngressExpiryInMs) {
+      this.log.warn('Timestamp is older than the max ingress expiry. Retrying query.', {
+        requestId,
+        timestamp: timestampInMs,
```
mraszyk: Could we move this to the subsequent if block? (We don't really retry if we already reached the maximum number of retries?

packages/agent/src/agent/http/index.ts
```diff
@@ -747,51 +732,35 @@ export class HttpAgent implements Agent {
...
-        new TimeoutWaitingForResponseErrorCode(
-          'Timestamp is below the watermark. Retrying query.',
-          requestId,
-        ),
-      );
-      this.log.error('Timestamp is below', error, {
-        timestamp,
-        waterMark: this.waterMark,
+    // If the timestamp is older than the max ingress expiry, retry the request up to the retry limit.
+    // We don't have to consider the #timeDiffMsecs here because the response comes from a single node,
```
mraszyk: I didn't get this comment? What is `#timeDiffMsecs` and why does it not matter?

e2e/node/utils/identity.ts
```diff
@@ -18,23 +18,23 @@ export interface KeyPair {
...
 
 /**
- * Generates a random key pair.
+ * Generates a random BLS key pair.
  * @returns {KeyPair} A new random key pair.
  */
 export function randomKeyPair(): KeyPair {
   const privateKey = bls.utils.randomPrivateKey();
-  const publicKey = bls.getPublicKeyForShortSignatures(privateKey);
+  const publicKey = bls.shortSignatures.getPublicKey(privateKey).toBytes(true);
```
nathanosdev, line 26: Was it causing an issue?

e2e/node/utils/mock-replica.ts
```diff
@@ -120,6 +156,9 @@ export class MockReplica {
         return;
       }
 
+      // add fallback implementation to return 500 if the spy runs out of implementations
+      spy.mockImplementation(fallbackSpyImpl(spyType, canisterId));
```
nathanosdev, line 161: Nice addition👌

packages/agent/src/agent/http/index.ts
ilbertt: Done

packages/agent/src/agent/http/index.ts
ilbertt: You're right, I've updated the error

packages/agent/src/agent/http/index.ts
nathanosdev: `#timeDiffMsecs` is the difference between the user's clock and the subnet's time (max of 3 replica's `time` path).

We should be taking that into account here or clock drift can be a problem.

e2e/node/utils/identity.ts
ilbertt: Not specifically, but it was marked as deprecated

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
ilbertt: Removed

packages/agent/src/agent/http/index.ts
```diff
@@ -750,11 +750,11 @@ export class HttpAgent implements Agent {
...
     const timestampInMs = Number(BigInt(timestampInNs) / BigInt(NANOSECONDS_TO_MSECS));
+    const ingressExpiryInMs =
+      this.#maxIngressExpiryInMs +
+      (mustReconcileTime(this.#timeDiffMsecs) ? this.#timeDiffMsecs : 0);
 
-    // If the timestamp is older than the max ingress expiry, retry the request up to the retry limit.
-    // We don't have to consider the #timeDiffMsecs here because the response comes from a single node,
-    // and hence it can be different from the network time.
-    if (Date.now() - timestampInMs > this.#maxIngressExpiryInMs) {
+    if (Math.abs(Date.now() - timestampInMs) > ingressExpiryInMs) {
```
mraszyk: I would avoid absolute value here (also see https://dfinity.atlassian.net/browse/SDK-2156).

packages/agent/src/agent/http/index.ts
```diff
@@ -750,11 +750,11 @@ export class HttpAgent implements Agent {
     }
 
     const timestampInMs = Number(BigInt(timestampInNs) / BigInt(NANOSECONDS_TO_MSECS));
+    const ingressExpiryInMs =
+      this.#maxIngressExpiryInMs +
+      (mustReconcileTime(this.#timeDiffMsecs) ? this.#timeDiffMsecs : 0);
```
mraszyk: The value of `this.#timeDiffMsecs` is [equal](https://github.com/dfinity/agent-js/blob/282fef9656d9fc7e0b22acf72602156fe7832458/packages/agent/src/agent/http/index.ts#L1307) to the difference `icpTimestamp - Date.now()`, where `icpTimestamp` is the trustworthy ICP time.

Hence, I find it more readable to define:
```
const icpTimestamp = Date.now() + this.#timeDiffMsecs
```
or
```
const icpTimestamp = Date.now() + (mustReconcileTime(this.#timeDiffMsecs) ? this.#timeDiffMsecs : 0)
```
(I don't see why we'd ignore low deltas here, but it's fine to do so.)

and then check
```
if (icpTimestamp - timestampInMs > this.#maxIngressExpiryInMs) {
```

packages/agent/src/agent/http/index.ts
```diff
@@ -1394,11 +1395,21 @@ export function calculateIngressExpiry(
...
       maxIngressExpiryInMinutes * MINUTE_TO_MSECS + timeDiffMsecs,
     );
   }
 
   return Expiry.fromDeltaInMilliseconds(maxIngressExpiryInMinutes * MINUTE_TO_MSECS);
 }
+
+/**
+ * Determines if the time should be reconciled with the network:
+ * if the time difference is more than 30 seconds, we need to reconcile the time.
```
mraszyk: Why do we only reconcile time if the time difference is more than 30 seconds? CC @nathanosdev

packages/agent/src/agent/http/index.ts
mraszyk: Now inlining `const icpTimestamp` yields
```
Date.now() + this.#timeDiffMsecs - timestampInMs > this.#maxIngressExpiryInMs
```
but the current check is
```
abs(Date.now() - timestampInMs) > this.#maxIngressExpiryInMs + this.#timeDiffMsecs
```
so for large values of `this.#timeDiffMsecs`, the current check is too weak (will miss errors): maybe test coverage could also be improved?

packages/agent/src/agent/http/index.ts
ilbertt: Removed

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 
 - feat: removes the watermark checks when checking query responses.
+- fix: always sync time with the IC
```
mraszyk: this is a bit misleading: the time is not always synced with the IC, but if it is, then the measured clock drift is always applied

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Changed
+
+- feat: removes the watermark checks when checking query responses.
```
mraszyk: Could we also say what we replaced the watermark check with? Now it sounds like we made the agent less secure.

packages/agent/src/agent/http/index.ts
```diff
@@ -81,7 +82,8 @@ export enum RequestStatusResponseStatus {
   Done = 'done',
 }
 
-const MINUTE_TO_MSECS = 60 * 1000;
+const MINUTE_TO_MSECS = 60 * 1_000;
+const NANOSECONDS_TO_MSECS = 1_000_000;
```
mraszyk:
```suggestion
const MSECS_TO_NANOSECONDS = 1_000_000;
```

packages/agent/src/agent/http/index.ts
```diff
@@ -1119,9 +1084,7 @@ export class HttpAgent implements Agent {
         request_type: ReadRequestType.ReadState,
         paths: fields.paths,
         sender,
-        ingress_expiry: Expiry.fromDeltaInMilliseconds(
-          this.#maxIngressExpiryInMinutes * MINUTE_TO_MSECS,
-        ),
+        ingress_expiry: Expiry.fromDeltaInMilliseconds(this.#maxIngressExpiryInMs),
```
mraszyk: We should use `calculateIngressExpiry` here, too.

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

CHANGELOG.md

CHANGELOG.md

PR #1048, opened 2025-07-03
chore: updates the version script to work with new pnpm workspaces

krpeacock:
# Description

Noticed this was broken while making the release today. It still doesn't account for a different version between core and other packages. Should we add that support while we're at it? We'll need to update the "prepare release" github action to have two inputs if we want to have multiple versions handled

### DIFF
--- a/bin/version.ts
+++ b/bin/version.ts
@@ -1,32 +1,35 @@
-// Bumps the version (@jsonpaths: .version, .dependencies, .peerDependencies, and .devDependencies
-// of each package's `package.json` in the monorepo.
+// Bumps the version (@jsonpaths: .version of each package's `package.json` in the monorepo.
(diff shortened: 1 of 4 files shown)


bin/version.ts
```diff
@@ -1,32 +1,35 @@
...
-const rootPackage = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, '..', 'package.json')).toString(),
+const workspaceConfig = yaml.parse(
+  fs.readFileSync(path.resolve(__dirname, '..', 'pnpm-workspace.yaml')).toString(),
 );
 
-if (!rootPackage.name) throw new Error("Couldn't find root package.json");
-
-const newVersion = process.argv[2];
+console.log(process.argv)
```
ilbertt: Can this be removed?

krpeacock:
```suggestion
```

PR #1051, opened 2025-07-10
docs: add initial starlight docs

nathanosdev:
# Description

This PR adds the initial docs implementation using Starlight. There's more work to do on this for sure, this is roughly equivalent to what we had before but looks nicer and sets us up for future iterations.

# Checklist:


packages/agent/src/certificate.ts
```diff
@@ -179,7 +179,6 @@ export class Certificate {
 
   /**
    * Create a new instance of a certificate, automatically verifying it.
-   * @constructs  Certificate
```
nathanosdev, line 182: Some changes are made to solve some TypeDoc warnings. There are loads more warnings still, I didn't wanna add too many changes to this PR.

docs/src/libs-plugin.ts
nathanosdev, line 1: This is the plugin. Everything is in one file for now, but this can be moved to another repo and made reusable.

The plugin does the following:

- Run TypeDoc for every package in the base dir.
  - Use sane defaults, including the markdown output plugin.
- Augment the TypeDoc markdown output (required for the markdown to work in Astro):
  - Add missing frontmatter to all markdown files.
  - Rewrite relative links to other markdown files.
- Add the readme for every package into the astro folder.
- Arrange all readme's together.
- Create the sidebar dynamically to provide entries for every package.

docs/astro.config.mjs
```diff
@@ -0,0 +1,28 @@
+// @ts-check
+import { defineConfig } from 'astro/config';
+import starlight from '@astrojs/starlight';
+import { libsPlugin } from './src/libs-plugin';
+
+// https://astro.build/config
+export default defineConfig({
+  site: 'https://dfinity.github.io/agent-js/',
```
ilbertt: Isn't https://agent-js.icp.xyz/ the actual website URL?

.github/workflows/publish.yml
```diff
@@ -44,8 +44,6 @@ jobs:
       # publish docs
       - name: Install dfx
         run: dfinity/setup-dfx@main
-      - name: Regenerate project's documentation
```
ilbertt, line 47: If we remove the docs build from the `pnpm build` command (see my review comment), we should add back the script to build the docs here

ilbertt (changes requested):
Great job!

I have a couple of points:

1. The docs/README.md files looks like the default one. Maybe we can add one intro line, remove everything except the _Commands_ section and add a little explanation of the Libs plugin?
2. Running `pnpm build` from the root of the repo triggers the build of the docs as well, which is not the behavior I would expect. I think building the docs should be kept separate from the actual packages builds.

.github/workflows/publish.yml
nathanosdev: I've left this as-is since `docs` wasn't removed from `pnpm build`. I added the `build:packages` script to make building without the docs easier.

PR #1055, opened 2025-07-16
fix: export ESM modules

ilbertt:
# Description

Changes the order in the `"exports"` fields in the packages' `package.json` file to prioritize ESM modules.
Adds a `package.json` inside each package's `lib/esm` output folder to tell any Node.js app that it's a module. 

# How Has This Been Tested?

Same tests should still pass.

# Checklist:

### DIFF
--- a/packages/agent/src/actor.test.ts
+++ b/packages/agent/src/actor.test.ts
@@ -337,6 +337,7 @@ describe('makeActor', () => {
     `);
     expect(replyUpdateWithHttpDetails.result).toEqual(canisterDecodedReturnValue);
 
+    // `requestDetails` is not inside the type. TODO: fix this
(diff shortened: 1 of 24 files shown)


packages/agent/src/actor.test.ts
```diff
@@ -337,7 +337,6 @@ describe('makeActor', () => {
     `);
     expect(replyUpdateWithHttpDetails.result).toEqual(canisterDecodedReturnValue);
 
-    // @ts-expect-error - `requestDetails` is not inside the type. TODO: fix this
```
ilbertt, line 340: I was getting an "Unexpected @ts-expected-error ..." error, so I removed it

nathanosdev: Let's keep the comment, but remove the `@ts-expect-error` part.

nathanosdev (approved):

PR #1056, opened 2025-07-16
chore: release notes for v3

ilbertt:
# Description

Adds release notes for v3 to the docs.

Additionally, updates the CHANGELOG.md.

Blocked by #1051.

# How Has This Been Tested?

Same tests should still pass.

# Checklist:

### DIFF
--- a/docs/src/content/docs/release-notes/v3.0.0.md
+++ b/docs/src/content/docs/release-notes/v3.0.0.md
@@ -2,3 +2,288 @@
 title: Release notes for v3.0.0
 description: Release notes for the ICP JavaScript SDK v3.0.0.
 ---
+
+This release marks a major update to the ICP JavaScript SDK, introducing several breaking changes, new features, and improvements. This guide is intended to help developers upgrade their projects to the latest version.
+
+## Breaking Changes
+
+### Removal of Management Canister from `@dfinity/agent`
+
+The management canister has been moved to its own package, `@dfinity/ic-management`. You will need to update your code to use this new package for management canister interactions.
+
+**Before:**
+
+```typescript
+import { getManagementCanister } from '@dfinity/agent';
+
+const management = getManagementCanister({ agent });
+```
+
+**After:**
+
+```typescript
+import { getManagementCanister } from '@dfinity/ic-management';
+
+const management = getManagementCanister({ agent });
+```
+
+You will also need to add `@dfinity/ic-management` as a dependency to your project.
+
+### Use `Uint8Array` instead of `ArrayBuffer`
+
+Internally, we have replaced the usage of `ArrayBuffer` with `Uint8Array`.
+
+Specifically, all `@dfinity/candid` interfaces that previously accepted or returned `ArrayBuffer` now use `Uint8Array`. You will need to update your code to handle `Uint8Array` instead of `ArrayBuffer`.
+
+**Before:**
+
+```typescript
+import { IDL } from '@dfinity/candid';
+
+const myArgType = IDL.Text;
+const myArgValue = 'Hello, world!';
+const encoded: ArrayBuffer = IDL.encode([myArgType], [myArgValue]);
+const decoded: [string] = IDL.decode([myArgType], encoded);
+```
+
+Or, with utility functions:
+
+```typescript
+import { concat } from '@dfinity/candid';
+
+const buffer1: ArrayBuffer = ...;
+const buffer2: ArrayBuffer = ...;
+const combined: ArrayBuffer = concat(buffer1, buffer2);
+```
+
+**After:**
+
+```typescript
+import { IDL } from '@dfinity/candid';
+
+const myArgType = IDL.Text;
+const myArgValue = 'Hello, world!';
+const encoded: Uint8Array = IDL.encode([myArgType], [myArgValue]);
+const decoded: [string] = IDL.decode([myArgType], encoded);
+```
+
+Or, with utility functions:
+
+```typescript
+import { concat } from '@dfinity/candid';
+
+const buffer1: Uint8Array = ...;
+const buffer2: Uint8Array = ...;
+const combined: Uint8Array = concat(buffer1, buffer2);
+```
+
+### Utility Function Replacements
+
+The `fromHex`, `toHex`, and `concat` utility functions have been replaced with `bytesToHex`, `hexToBytes`, and `concatBytes` from `@noble/hashes/utils`.
+
+**Before:**
+
+```typescript
+import { toHex, fromHex, concat } from '@dfinity/agent';
+
+const hex = toHex(myUint8Array);
+const bytes = fromHex(myHexString);
+const combined = concat(buffer1, buffer2);
+```
+
+**After:**
+
+```typescript
+import { bytesToHex, hexToBytes, concatBytes } from '@noble/hashes/utils';
+
+const hex = bytesToHex(myUint8Array);
+const bytes = hexToBytes(myHexString);
+const combined = concatBytes(buffer1, buffer2);
+```
+
+### New `AgentError` Error and Better Error Handling
+
+Several specific error classes have been removed in favor of a new, more generic `AgentError`. This new error class has the following properties:
+
+- `code`: an `ErrorCode` class value, e.g. `CertifiedRejectErrorCode`. See [`errors.ts`](https://github.com/dfinity/agent-js/blob/e7e371b5399d44ea387f21cdcb0e195af4bef2b9/packages/agent/src/errors.ts) for all error codes.
+- `kind`: an `ErrorKindEnum` enum value, e.g. `ErrorKindEnum.Trust`. See [`ErrorKindEnum`](https://github.com/dfinity/agent-js/blob/e7e371b5399d44ea387f21cdcb0e195af4bef2b9/packages/agent/src/errors.ts#L9-L18) for all error kinds.
+- [`message`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message): a human-readable error message
+- [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause): an object containing the `code` and `kind` properties
+
+This new error class is more flexible and easier to handle in your code.
+
+The removed error classes are:
+
+- `AgentHTTPResponseError`
+- `AgentCallError`
+- `AgentQueryError`
+- `AgentReadStateError`
+- `CertificateVerificationError`
+- `ActorCallError`
+- `QueryCallRejectedError`
+- `UpdateCallRejectedError`
+
+You should update your error handling logic to catch `AgentError` and inspect its properties.
+
+**Example:**
+
+```typescript
+try {
+  // agent call
+} catch (e) {
+  if (e instanceof AgentError) {
+    console.error(e.message);
+    if (
+      e.code instanceof CertifiedRejectErrorCode && // or e.cause.code
+      e.kind === ErrorKindEnum.Trust // or e.cause.kind
+    ) {
+      // do something
+    }
+  }
+}
+```
+
+### Removal of `defaultAgent`
+
+The global `defaultAgent` concept and the `getDefaultAgent` function have been removed. The `HttpAgent` constructor is now the only way to create an agent.
+
+**Before:**
+
+```typescript
+import { getDefaultAgent } from '@dfinity/agent';
+
+const agent = getDefaultAgent();
+```
+
+**After:**
+
+```typescript
+import { HttpAgent } from '@dfinity/agent';
+
+const agent = await HttpAgent.create(); // or use the synchronous version `HttpAgent.createSync()`
+```
+
+### Removal of `ProxyAgent`
+
+The `ProxyAgent` class has been removed. You should use the `HttpAgent` class instead.
+
+### `Expiry` Class Refactor
+
+The `Expiry` class has been refactored to use static factory methods and no longer exposes a public constructor. It also now supports JSON serialization/deserialization.
+
+**Before:**
+
+```typescript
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
 description: Release notes for the ICP JavaScript SDK v3.0.0.
 ---
+
+This release marks a major update to the ICP JavaScript SDK, introducing several breaking changes, new features, and improvements. This guide is intended to help developers upgrade their projects to the latest version.
+
+## Breaking Changes
+
+### Removal of Management Canister from `@dfinity/agent`
+
+The management canister has been moved to its own package, `@dfinity/ic-management`. You will need to update your code to use this new package for management canister interactions.
```
peterpeterparker: `@dfinity/ic-management` has been available for ages. If you want to be accurate you could mention that agent-js implementation was a duplicate.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+```typescript
+import { getManagementCanister } from '@dfinity/agent';
+
+const management = getManagementCanister({ agent });
+```
+
+**After:**
+
+```typescript
+import { getManagementCanister } from '@dfinity/ic-management';
```
peterpeterparker: This is incorrect, where did you find the snippet? We need to fix it.

see correct usage there -> https://github.com/dfinity/ic-js/tree/main/packages/ic-management#usage

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+
+const management = getManagementCanister({ agent });
+```
+
+You will also need to add `@dfinity/ic-management` as a dependency to your project.
```
peterpeterparker: The package isn’t compatible with Agent-js v3. Are you sharing this changelog in advance and waiting for the release, or do you plan to publish it despite the incompatibility? If it’s the latter, it might be worth rephrasing or adding a note about this limitation.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+
+### Use `Uint8Array` instead of `ArrayBuffer`
+
+Internally, we have replaced the usage of `ArrayBuffer` with `Uint8Array`.
```
peterpeterparker, line 59: Does it actually require the consumer to upgrade TypeScript to a specific version?

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
 title: Release notes for v3.0.0
 description: Release notes for the ICP JavaScript SDK v3.0.0.
 ---
+
+This release marks a major update to the ICP JavaScript SDK, introducing several breaking changes, new features, and improvements. This guide is intended to help developers upgrade their projects to the latest version.
```
peterpeterparker: I would add a brief sentence mentioning that it eliminates Buffer and similar dependencies, reducing bundle size and improving performance — like an appetizer to encourage the reader to continue.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+import { concat } from '@dfinity/candid';
+
+const buffer1: Uint8Array = ...;
+const buffer2: Uint8Array = ...;
+const combined: Uint8Array = concat(buffer1, buffer2);
+```
+
+### Utility Function Replacements
+
+The `fromHex`, `toHex`, and `concat` utility functions have been replaced with `bytesToHex`, `hexToBytes`, and `concatBytes` from `@noble/hashes/utils`.
```
peterpeterparker: "replaced" or "renamed" - i.e. just refactoring or reimplemented? In case solely refactored, maybe "renamed" is more accurate.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+const agent = await HttpAgent.create(); // or use the synchronous version `HttpAgent.createSync()`
+```
+
+### Removal of `ProxyAgent`
+
+The `ProxyAgent` class has been removed. You should use the `HttpAgent` class instead.
+
+### `Expiry` Class Refactor
+
+The `Expiry` class has been refactored to use static factory methods and no longer exposes a public constructor. It also now supports JSON serialization/deserialization.
```
peterpeterparker, line 197: I doubt anyone besides me would care or that it would conflict with consumer implementations, but for accuracy, I was wondering if we should add a note that the values are serialized under an __expiry__ key or something?

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+  certificate.tree,
+);
+
+// Lookup a subtree, with the new lookup_subtree method
+const subnetLookupResult = certificate.lookup_subtree(['subnet', delegation.subnet_id, 'node']);
+```
+
+### Hashing Function Replacement
+
+The `hash` function has been replaced with `sha256` from `@noble/hashes/sha2`.
```
peterpeterparker: Same idea as my previous comment: "renamed" or "replaced" here?

(Just asking in the spirit of clarity, it already looks really good.)

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+
+```typescript
+import { sha256 } from '@noble/hashes/sha2';
+
+const hashed = sha256(myBuffer);
+```
+
+### Cbor package replacement
+
+The `@dfinity/agent` package now uses [`@dfinity/cbor`](https://npmjs.com/package/@dfinity/cbor) instead of the `borc` and `simple-cbor` dependencies, reducing [a lot](https://github.com/dfinity/agent-js/pull/1015#issuecomment-2912381841) the bundle size.
```
peterpeterparker: > [a lot]

I would throw in a number and emoji, like “by 30% gzipped 🤯,” to make it stand out. It’s awesome!

I also wonder if that shouldn’t be the first feature listed, by the way.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,288 @@
...
+
+### Node.js Version Support
+
+Support for Node.js v19 and lower, and Node.js v21 has been dropped. Please ensure you are using a supported version of Node.js (LTS versions like 20 or 22 are recommended).
+
+## New Features
+
+- **Improved Polling Strategy**: The polling strategy for `read_state` requests has been changed to support presigned requests. A new `preSignReadStateRequest` option allows for a single signature to be used for all polling requests, which is beneficial for hardware wallets.
+
+- **Watermark Check for Query Responses**: The agent now checks if the node signature is not older than the `ingressExpiryInMinutes` option, taking into account clock drift, to prevent stale query responses.
```
peterpeterparker: The "five-minute offset in the past or future" has often been an issue for many developers and users, though it seems better those days. Therefore, I was wondering if this change should not be highlighted more prominently and described in a bit more detail? Or maybe that would be too technical for this context?

docs/src/content/docs/release-notes/v3.0.0.md
ilbertt: No, it shouldn't

docs/src/content/docs/release-notes/v3.0.0.md
ilbertt: I think the JSDoc on the `Expiry.toJSON` method is enough in this case.

docs/src/content/docs/release-notes/v3.0.0.md
ilbertt: Maybe it's too technical. By the way, I've moved it to the "Fixes and Improvements", which sounds more appropriate.

docs/src/content/docs/release-notes/v3.0.0.md
ilbertt: I've rephrased the sentence

docs/src/content/docs/release-notes/v3.0.0.md
ilbertt: Same as above, I've rephrased the sentence

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+
+```typescript
+import { HttpAgent } from '@dfinity/agent';
+
+const agent = await HttpAgent.create(); // or use the synchronous version `HttpAgent.createSync()`
+```
+
+### Removal of `ProxyAgent`
+
+The `ProxyAgent` class has been removed. You should use the `HttpAgent` class instead.
```
nathanosdev: The `HttpAgent` is not a replacement for `ProxyAgent` so this suggestion doesn't make much sense. I'm not sure if anyone was using that, but if they are and they still need it then the solution would be to roll their own `ProxyAgent` that uses `HttpAgent` internally.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+
+The `Expiry` class has been refactored to use static factory methods and no longer exposes a public constructor. It also now supports JSON serialization/deserialization.
+
+**Before:**
+
+```typescript
+import { Expiry } from '@dfinity/agent';
+
+// This is a simplified example of what might have been done,
+// though direct instantiation was often internal.
```
nathanosdev: I'm not sure what `though direct instantiation was often internal` means.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+const expiry = Expiry.fromDeltaInMilliseconds(fiveMinutes);
+
+// You can also serialize and deserialize Expiry objects.
+const json = expiry.toJSON();
+const expiryFromJson = Expiry.fromJSON(JSON.stringify(json));
+```
+
+### Certificate Method Renaming and Removal
+
+- The `lookup` method of the `Certificate` class has been renamed to `lookup_path` to align with the [IC Interface Specification](https://github.com/dfinity/portal/blob/8015a4ab50232176723ffd95e32a02f1bf7fef30/docs/references/ic-interface-spec.md?plain=1#L3069).
```
nathanosdev: I believe `lookup` was renamed to `lookup_subtree` and `lookup_path` is the new method that matches the spec. The old method didn't match the spec.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+```
+
+**After:**
+
+```typescript
+// Lookup a path
+const result = certificate.lookup_path(['a', 'b']);
+const label = certificate.lookup_path(['path', 'to', 'label']);
+
+// Additionally, you can now lookup subtrees.
```
nathanosdev: It was already the case before that you could lookup subtrees with `lookup`.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+  certificate.tree,
+);
+
+// Lookup a subtree, with the new lookup_subtree method
+const subnetLookupResult = certificate.lookup_subtree(['subnet', delegation.subnet_id, 'node']);
+```
+
+### Hashing Function Replacement
+
+The `hash` function has been removed. Use the `sha256` function from `@noble/hashes/sha2` instead.
```
nathanosdev: Maybe a better recommendation is the native [digest()](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) method, with `sha256` as a synchronous alternative.

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+| ----------------------------- | ---------------- | --------------- | ------------------ |
+| `@dfinity/identity-secp256k1` | 106.69 kB        | 32.95 kB        | - 69% 🔽           |
+| `@dfinity/identity`           | 47.8 kB          | 18.76 kB        | - 61% 🔽           |
+| `@dfinity/auth-client`        | 50.15 kB         | 19.4 kB         | - 61% 🔽           |
+| `@dfinity/agent`              | 72.1 kB          | 47.45 kB        | - 34% 🔽           |
+| `@dfinity/assets`             | 67.97 kB         | 52.51 kB        | - 22% 🔽           |
+| `@dfinity/principal`          | 4.21 kB          | 4.44 kB         | + 5% 🔺            |
+| `@dfinity/candid`             | 11.82 kB         | 12.96 kB        | + 10% 🔺           |
+
+### Clock Drift Calculation
```
nathanosdev, line 331: I think it's worth mentioning the altered behavior for fetch root key and sync time.

The deferred method works with the deprecated constructor as well, which I guess is worth mentioning somehow.

**Deferred method**
```
HttpAgent.createSync({ shouldSyncTime: true, fetchRootKey: true }); 

await actor.canister_method(); // waits here for sync time and fetch root key
```

**Non-deferred method**
```
await HttpAgent.create(); // waits here for sync time and fetch root key
```

docs/src/content/docs/release-notes/v3.0.0.md
```diff
@@ -2,3 +2,313 @@
...
+
+**Example:**
+
+```typescript
+const agent = await HttpAgent.create({
+  shouldSyncTime: true,
+});
+```
+
+### No more watermark checks for query responses
```
nathanosdev, line 361: It may be worth mentioning that this should get rid the watermark errors that were frequently reported on mainnet.

nathanosdev (approved):

PR #1076, opened 2025-07-23
fix: round ingress expiry before applying clock drift

ilbertt:
# Description

Calculates the ingress expiry by first rounding down the timeout and only then applying the clock drift.

As a consequence, adds the `deltaInMsAfterThresholdCheck` optional parameter to `Expiry.fromDeltaInMilliseconds` to allow for adding a delta after the internal threshold check.

Blocked by #1075.

# How Has This Been Tested?

Same tests should still pass. Added unit tests for `addMilliseconds`.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## [Unreleased]
 
 - fix: do not subtract the replica permitted clock drift when calculating the ingress expiry.
-- fix: rounds the ingress expiry before applying any eventual clock drift.
-- feat: adds a `deltaInMsAfterThresholdCheck` optional parameter to `Expiry.fromDeltaInMilliseconds` to allow for adding a delta after the internal threshold check.
+- fix: pick the expiry rounding strategy based on the delta, without adding the clock drift to the delta.
+- feat: adds a `clockDriftMs` optional parameter to `Expiry.fromDeltaInMilliseconds` to add to the current time, typically used to specify the clock drift between the client's clock and the IC network clock.
 
 ## [3.1.0] - 2025-07-24
 
--- a/packages/agent/src/agent/http/transforms.ts
+++ b/packages/agent/src/agent/http/transforms.ts
@@ -38,22 +38,18 @@ export class Expiry {
    * If the delta is less than 90 seconds, the expiry is rounded down to the nearest second.
    * Otherwise, the expiry is rounded down to the nearest minute.
    * @param deltaInMs The milliseconds to add to the current time.
-   * @param deltaInMsAfterThresholdCheck The milliseconds to add to the current time after the 90 seconds threshold check. Defaults to `0` if not provided.
+   * @param clockDriftMs The milliseconds to add to the current time, typically the clock drift between the client and the IC network clock. Defaults to `0` if not provided.
    * @returns {Expiry} The constructed Expiry object.
    */
-  public static fromDeltaInMilliseconds(
-    deltaInMs: number,
-    deltaInMsAfterThresholdCheck: number = 0,
-  ): Expiry {
+  public static fromDeltaInMilliseconds(deltaInMs: number, clockDriftMs: number = 0): Expiry {
     const deltaMs = BigInt(deltaInMs);
-    const deltaMsAfterThresholdCheck = BigInt(deltaInMsAfterThresholdCheck);
-    const expiryMs = BigInt(Date.now()) + deltaMs;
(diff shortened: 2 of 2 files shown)


packages/agent/src/agent/http/index.ts
```diff
@@ -1398,7 +1398,6 @@ export function calculateIngressExpiry(
   maxIngressExpiryInMinutes: number,
   timeDiffMsecs: number,
 ): Expiry {
-  return Expiry.fromDeltaInMilliseconds(
-    maxIngressExpiryInMinutes * MINUTE_TO_MSECS + timeDiffMsecs,
-  );
+  const roundedExpiry = Expiry.fromDeltaInMilliseconds(maxIngressExpiryInMinutes * MINUTE_TO_MSECS);
+  return roundedExpiry.addMilliseconds(timeDiffMsecs);
```
mraszyk: I'd still round at the very end for the sake of caching, but make sure the rounding strategy (round to seconds vs round to minutes) is based on `maxIngressExpiryInMinutes` only, i.e., not affected by `timeDiffMsecs`.

ilbertt: Fixed!

packages/agent/src/agent/http/transforms.ts
```diff
@@ -37,18 +37,23 @@ export class Expiry {
    * Creates an Expiry object from a delta in milliseconds.
    * If the delta is less than 90 seconds, the expiry is rounded down to the nearest second.
    * Otherwise, the expiry is rounded down to the nearest minute.
-   * @param deltaInMs The delta in milliseconds.
+   * @param deltaInMs The milliseconds to add to the current time.
+   * @param deltaInMsAfterThresholdCheck The milliseconds to add to the current time after the 90 seconds threshold check. Defaults to `0` if not provided.
```
mraszyk: I'd rename this variable so that its intent is more clear, namely to account for the clock drift. Then the fact that we don't consider it in the threshold check is merely a consequence (now somebody reading this code might wonder why the deltas are split and how to set those parameters).

packages/agent/src/agent/http/transforms.ts
```diff
@@ -37,18 +37,23 @@ export class Expiry {
...
    * @returns {Expiry} The constructed Expiry object.
    */
-  public static fromDeltaInMilliseconds(deltaInMs: number): Expiry {
+  public static fromDeltaInMilliseconds(
+    deltaInMs: number,
+    deltaInMsAfterThresholdCheck: number = 0,
+  ): Expiry {
     const deltaMs = BigInt(deltaInMs);
+    const deltaMsAfterThresholdCheck = BigInt(deltaInMsAfterThresholdCheck);
     const expiryMs = BigInt(Date.now()) + deltaMs;
```
mraszyk:
```suggestion
    const expiryMs = BigInt(Date.now()) + deltaMs + deltaMsAfterThresholdCheck;
```

mraszyk: and then we could drop it from both if-branches below

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 - fix: do not subtract the replica permitted clock drift when calculating the ingress expiry.
+- fix: rounds the ingress expiry before applying any eventual clock drift.
```
mraszyk:
```suggestion
```
no longer the case

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
...
+- feat: adds a `deltaInMsAfterThresholdCheck` optional parameter to `Expiry.fromDeltaInMilliseconds` to allow for adding a delta after the internal threshold check.
```
mraszyk: I'd also point out the intent (clock drift) of this parameter in the changelog (and ideally also rename it as pointed out in a separate comment).

mraszyk (commented):
LGTM semantically, left a few minor comments about naming and documentation

CHANGELOG.md

CHANGELOG.md
ilbertt: Instead of removing it, I've adjusted the wording

packages/agent/src/agent/http/transforms.ts
ilbertt: Fixed

packages/agent/src/agent/http/transforms.ts
ilbertt: Fixed

mraszyk (approved):

PR #1081, opened 2025-07-25
fix: account for clock drift in certificate freshness check

### LINKED ISSUE #686: Correct error if wasm for BLS signature verification can't be instantiated
Currently, an error thrown by the BLS verification gets caught and a very generic [`Signature verification failed` is thrown](https://github.com/dfinity/agent-js/blob/aa14db37898852cb7293aeda67ed09dc9ea61a91/packages/agent/src/certificate.ts#L179). In most cases, the reason for failure is that the IC public key is used though the message is signed by the local replica. However, in my case, a CSP directive prevented the loading of the wasm, but since this error was caught it took me quite a while to discover this reason:

```
CompileError: WebAssembly.instantiate(): Refused to compile or instantiate WebAssembly module because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https://www.google-analytics.com https://www.googletagmanager.com 'unsafe-inline'"
```

It would be great if we would distinguish between this error and the typical failure to verify the signature.

ilbertt:
# Description

Accounts for the clock drift between the current time and the IC network time when checking the certificate freshness. Adds some helper methods and functions, check the updated changelog for a list of all the changes.

Additionally, fixes #686.

# How Has This Been Tested?

Added unit tests to check the new clock drift adjustments in the certificate verification.

# Checklist:


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - fix: do not subtract the replica permitted clock drift when calculating the ingress expiry.
 - fix: pick the expiry rounding strategy based on the delta, without adding the clock drift to the delta.
 - feat: adds a `clockDriftMs` optional parameter to `Expiry.fromDeltaInMilliseconds` to add to the current time, typically used to specify the clock drift between the client's clock and the IC network clock.
+- fix: account for clock drift when verifying the certificate freshness.
+- feat: adds the `timeDiffMsecs` optional field to the `CreateCertificateOptions` interface, which allows you to adjust the current time when verifying the certificate freshness.
+- feat: adds the `getTimeDiffMsecs` function to the `HttpAgent` class, which returns the time difference in milliseconds between the client's clock and the IC network clock. It also adds the `getTimeDiffMsecs` function to handle the case where the `HttpAgent` is not an instance of `HttpAgent`.
```
mraszyk: > where the `HttpAgent` is not an instance of `HttpAgent`.

sounds contradictory

packages/agent/src/errors.ts
```diff
@@ -208,14 +219,15 @@ export class CertificateTimeErrorCode extends ErrorCode {
     public readonly maxAgeInMinutes: number,
     public readonly certificateTime: Date,
     public readonly currentTime: Date,
+    public readonly timeDiffMsecs: number,
     public readonly ageType: 'past' | 'future',
```
mraszyk, line 223: Observing a certificate in the future hints at the client's clock being out of sync - see another comment next to the corresponding check.

packages/agent/src/agent/http/index.ts
```diff
@@ -1386,6 +1386,15 @@ export class HttpAgent implements Agent {
 
     return p;
   }
+
+  /**
+   * Returns the time difference in milliseconds between the client's clock and the IC network clock,
+   * after the clock has been synced using the {@link HttpAgent.syncTime} method
+   * or during agent creation if {@link HttpAgentOptions.shouldSyncTime} was set to `true`.
+   */
```
mraszyk: Otherwise, it returns 0, right? Or it is an optional value (better)?

packages/agent/src/agent/http/index.ts
```diff
@@ -1402,3 +1411,17 @@ export function calculateIngressExpiry(
   const ingressExpiryMs = maxIngressExpiryInMinutes * MINUTE_TO_MSECS;
   return Expiry.fromDeltaInMilliseconds(ingressExpiryMs, timeDiffMsecs);
 }
+
+/**
+ * Retrieves the time different in milliseconds between the client's clock and the IC network clock.
```
mraszyk:
```suggestion
 * Retrieves the time difference in milliseconds between the client's clock and the IC network clock.
```

packages/agent/src/agent/http/index.ts
```diff
@@ -1402,3 +1411,17 @@ export function calculateIngressExpiry(
...
+ * See {@link HttpAgent.getTimeDiffMsecs} for more details.
+ * @param agent The agent to retrieve the `timeDiffMsecs` property from.
+ * @returns The time different in milliseconds between the client's clock and the IC network clock,
```
mraszyk:
```suggestion
 * @returns The time difference in milliseconds between the client's clock and the IC network clock,
```

packages/agent/src/certificate.ts
```diff
@@ -253,39 +258,47 @@ export class Certificate {
...
     if (!this.#disableTimeVerification) {
-      const FIVE_MINUTES_IN_MSEC = 5 * 60 * 1000;
-      const MAX_AGE_IN_MSEC = this._maxAgeInMinutes * 60 * 1000;
+      const maxAgeInMsec = this._maxAgeInMinutes * MINUTES_TO_MSEC;
       const now = Date.now();
-      const earliestCertificateTime = now - MAX_AGE_IN_MSEC;
-      const fiveMinutesFromNow = now + FIVE_MINUTES_IN_MSEC;
+      const adjustedNow = now + timeDiffMsecs;
+      const earliestCertificateTime = adjustedNow - maxAgeInMsec;
+      const latestCertificateTime = adjustedNow + FIVE_MINUTES_IN_MSEC;
```
mraszyk, line 277: If this check fails (certificate time is in the future), then clock is not synchronized so we should sync time and retry the check.

packages/agent/src/certificate.ts
```diff
@@ -180,16 +191,12 @@ export class Certificate {
...
-   * @param {Principal} options.canisterId The effective or signing canister ID
-   * @param {number} options.maxAgeInMinutes The maximum age of the certificate in minutes. Default is 5 minutes.
-   * @throws if the certificate cannot be verified
+   * @throws if the verification of the certificate fails
    */
   public static async create(options: CreateCertificateOptions): Promise<Certificate> {
     const cert = Certificate.createUnverified(options);
 
-    await cert.verify();
+    await cert.verify(options.timeDiffMsecs ?? 0);
```
mraszyk: If timeDiffMsecs is not set and this fails, then we should sync time and retry.

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
ilbertt: It returns `0`, which is the default when the agent is initialized. I've updated the doc comment to be more clear.

CHANGELOG.md

packages/agent/src/certificate.ts
ilbertt: This will be addressed in https://github.com/dfinity/agent-js/pull/1082

packages/agent/src/certificate.ts
ilbertt: This will be addressed in https://github.com/dfinity/agent-js/pull/1082

packages/agent/src/errors.ts
ilbertt: This will be addressed in https://github.com/dfinity/agent-js/pull/1082

packages/agent/src/agent/http/index.ts
```diff
@@ -1402,3 +1413,17 @@ export function calculateIngressExpiry(
...
 }
+
+/**
+ * Retrieves the time difference in milliseconds between the client's clock and the IC network clock.
+ * See {@link HttpAgent.getTimeDiffMsecs} for more details.
+ * @param agent The agent to retrieve the `timeDiffMsecs` property from.
+ * @returns The time difference in milliseconds between the client's clock and the IC network clock,
+ * if the agent is an {@link HttpAgent} instance. `undefined` otherwise.
+ */
+export function getTimeDiffMsecs(agent: Agent | HttpAgent): number | undefined {
```
nathanosdev: Why is this additional function necessary? Why not call `agent.getTimeDiffMsecs` instead of `getTimeDiffMsecs(agent)`?

ilbertt: If the agent is not an instance of `HttpAgent`, the `getTimeDiffMsecs` method does not exist. This is a convenience function to do that check

nathanosdev (approved):

packages/agent/src/certificate.ts
```diff
@@ -271,14 +269,20 @@ export class Certificate {
           new CertificateTimeErrorCode(
             this._maxAgeInMinutes,
             certTime,
-            new Date(now),
-            timeDiffMsecs,
+            new Date(),
+            currentTimestamp - Date.now(),
```
mraszyk: These `new Date()` and `Date.now` are different from that used in the certificate freshness check: if we need to pass those values, we should pass the same values as used in the checks.

packages/agent/src/certificate.ts
```diff
@@ -271,14 +269,20 @@ export class Certificate {
...
           ),
         );
       } else if (certTime.getTime() > latestCertificateTime) {
         throw TrustError.fromCode(
-          new CertificateTimeErrorCode(5, certTime, new Date(now), timeDiffMsecs, 'future'),
+          new CertificateTimeErrorCode(
+            5,
+            certTime,
+            new Date(),
+            currentTimestamp - Date.now(),
```
mraszyk: same here

packages/agent/src/certificate.ts
mraszyk: I'm not sure #1082 is sufficient to address this since that PR only takes care of certificates in canister status request, not other kinds of certificates.

packages/agent/src/certificate.ts
ilbertt: The idea here is to show an error message with the current system time and the eventual clock drift, to make it clearer to devs what is happening

packages/agent/src/certificate.ts
ilbertt: Yeah, #1082 needs more work

mraszyk: I think #1082 is fine, I'd create a separate PR for the follow up.

packages/agent/src/certificate.ts
mraszyk: Makes sense, in that case, I'd suggest to use the same values in the error message as used during the check, otherwise it could be also confusing.

ilbertt: I've pushed an update for this

packages/agent/src/errors.ts
```diff
@@ -215,19 +215,23 @@ export class CertificateVerificationErrorCode extends ErrorCode {
...
   constructor(
     public readonly maxAgeInMinutes: number,
     public readonly certificateTime: Date,
     public readonly currentTime: Date,
-    public readonly timeDiffMsecs: number,
     public readonly ageType: 'past' | 'future',
   ) {
     super();
     Object.setPrototypeOf(this, CertificateTimeErrorCode.prototype);
+    this.systemTime = new Date();
```
mraszyk: Why don't we take the system time from which the "adjusted" current time was derived?

ilbertt: Yeah, that would be the best solution and would make the error print the proper clock drift, but I can't think of a good way to pass it to the certificate without introducing a breaking change or making the options less clear. Maybe I should just revert https://github.com/dfinity/agent-js/pull/1081/commits/9f8ed1f71e69eaffcce225f7f7a610388c62dc73?

mraszyk: > Maybe I should just revert https://github.com/dfinity/agent-js/commit/9f8ed1f71e69eaffcce225f7f7a610388c62dc73?

Reverting would mean we don't touch this piece of code (so it does not attract attention on this PR), but the code already exhibited the exact same issue before this PR. So if there's no clear way forward at this point, let's just create a ticket about this issue and resolve this conversation.

mraszyk: Or would it be possible to clean up this piece of code after reverting? (Then I'd be up for reverting and _following up_.)

packages/agent/src/certificate.ts
ilbertt: I've updated this branch to sync time if the certificate is not fresh

packages/agent/src/certificate.ts
ilbertt: See https://github.com/dfinity/agent-js/pull/1081#discussion_r2251508151

packages/agent/src/errors.ts
ilbertt: See https://github.com/dfinity/agent-js/pull/1081#discussion_r2251508151

packages/agent/src/agent/http/index.ts
```diff
@@ -181,6 +183,15 @@ export interface HttpAgentOptions {
    * Whether or not to sync the time with the network during construction. Defaults to false.
    */
   shouldSyncTime?: boolean;
+
+  /**
+   * The time difference in milliseconds between the client's clock and the IC network clock.
```
mraszyk:
```suggestion
   * The time difference in milliseconds between the IC network clock and the client's clock.
```
it is set to `maxReplicaTime - callTime` in code

packages/agent/src/agent/http/index.ts
```diff
@@ -181,6 +183,15 @@ export interface HttpAgentOptions {
...
+   * This is used to adjust the current time when verifying the certificate freshness.
+   * If a value > `0` is provided, the {@link HttpAgent.syncTime} method will not be called during construction,
+   * even if {@link HttpAgentOptions.shouldSyncTime} is set to `true`.
+   * @default 0
```
mraszyk: could we replace `0` by `DEFAULT_TIME_DIFF_MSECS`?

packages/agent/src/agent/http/index.ts
```diff
@@ -181,6 +183,15 @@ export interface HttpAgentOptions {
    * Whether or not to sync the time with the network during construction. Defaults to false.
    */
   shouldSyncTime?: boolean;
+
+  /**
+   * The time difference in milliseconds between the client's clock and the IC network clock.
+   * This is used to adjust the current time when verifying the certificate freshness.
+   * If a value > `0` is provided, the {@link HttpAgent.syncTime} method will not be called during construction,
```
mraszyk: instead of > we should check for !=

packages/agent/src/agent/http/index.ts
```diff
@@ -949,7 +961,13 @@ export class HttpAgent implements Agent {
     // Attempt to make the query i=retryTimes times
     // Make query and fetch subnet keys in parallel
     try {
-      const [queryResult, subnetStatus] = await Promise.all([makeQuery(), getSubnetStatus()]);
+      const [queryResult, subnetStatus] = await Promise.all([
+        makeQuery(),
+        getSubnetStatus().catch(err => {
+          this.log.warn('Failed to fetch subnet keys. Error:', err);
+          return undefined;
```
mraszyk, line 959: Shouldn't we rethrow the error here?

packages/agent/src/agent/http/index.ts
```diff
@@ -1386,6 +1404,25 @@ export class HttpAgent implements Agent {
...
+   */
+  public getTimeDiffMsecs(): number {
+    return this.#timeDiffMsecs;
+  }
+
+  /**
+   * Returns `true` if the time has been synced at least once with the IC network, `false` otherwise.
+   */
+  public hasSyncedTime(): boolean {
+    // It's really unlikely that the clock drift is still 0 after the time has been synced
```
mraszyk: It is still possible though - it is difficult to use a private Boolean field for that?

packages/agent/src/agent/http/index.ts
ilbertt: We could, but since `DEFAULT_TIME_DIFF_MSECS` is not public, the published docs on the website wouldn't show its actual value

packages/agent/src/agent/http/index.ts
ilbertt: Improved

packages/agent/src/agent/http/index.ts
ilbertt: Fixed

packages/agent/src/agent/http/index.ts
mraszyk: Do we also check for != in the actual code?

packages/agent/src/agent/http/index.ts
ilbertt: Improved

packages/agent/src/agent/http/index.ts
```diff
@@ -1386,6 +1406,24 @@ export class HttpAgent implements Agent {
 
     return p;
   }
+
+  /**
+   * Returns the time difference in milliseconds between the client's clock and the IC network clock,
+   * after the clock has been synced using the {@link HttpAgent.syncTime} method
+   * or during agent creation if {@link HttpAgentOptions.shouldSyncTime} was set to `true`.
```
mraszyk:
```suggestion
   * after the clock has been synced.
```

I wouldn't go into details when time is synced since the details might soon get outdated.

packages/agent/src/agent/http/index.ts
nathanosdev: In the case that it fails, it will eventually hit this line of code and throw: https://github.com/dfinity/agent-js/blob/de5e3afa9c851c1f621a23f85d61918b83988fd3/packages/agent/src/agent/http/index.ts#L1020-L1022

@ilbertt it would be nice to make that param in `#verifyQueryResponse` mandatory and throw the exception earlier in my opinion.

ilbertt: The error is thrown in the `#verifyQuerySignature` function: https://github.com/dfinity/agent-js/blob/de5e3afa9c851c1f621a23f85d61918b83988fd3/packages/agent/src/agent/http/index.ts#L1021. Does it make sense?

Throwing an error in the `Promise.all` causes the other in-flight promises to be cancelled, which may not be what we want here

ilbertt: Fixed in #1091

packages/agent/src/agent/http/index.ts
```diff
@@ -1386,6 +1396,23 @@ export class HttpAgent implements Agent {
 
     return p;
   }
+
+  /**
+   * Returns the time difference in milliseconds between the client's clock and the IC network clock,
```
mraszyk:
```suggestion
   * Returns the time difference in milliseconds between the IC network clock and the client's clock,
```

mraszyk (approved):

PR #1085, opened 2025-07-30
docs: move docs to js.icp.build/core

ilbertt:
# Description

Move docs under the new js.icp.build domain, in the /core subpath. The new website is deployed using [Juno](https://juno.build).

Preview: https://xavwt-jqaaa-aaaal-asjaa-cai.icp0.io/core

# How Has This Been Tested?

Same tests should still pass.

# Checklist:


dfx.json
```diff
@@ -1,10 +0,0 @@
-{
```
nathanosdev, line 1: Is it not still useful for local testing or do we use Juno? Maybe some small addition to the contributing guide would be nice for that.

docs/astro.config.mjs
```diff
@@ -5,13 +5,15 @@ import { libsPlugin } from './src/libs-plugin';
...
-  site: 'https://agent-js.icp.xyz/',
+  site: 'https://js.icp.build/',
+  base: '/core/',
   integrations: [
     starlight({
       title: 'ICP JS SDK Docs',
       social: [{ icon: 'github', label: 'GitHub', href: 'https://github.com/dfinity/agent-js' }],
       plugins: [
         libsPlugin({
+          clean: true,
```
nathanosdev, line 16: This should default to `true` anyway, was that not working as expected?

docs/src/libs-plugin.ts
```diff
@@ -108,14 +108,16 @@ const prettyUrlsPlugin: RemarkPlugin =
...
       }
 
       // skip any other full URLs
       if (
         url.startsWith('https://') ||
         url.startsWith('/') ||
         url.startsWith('http://') ||
-        url.startsWith('mailto:')
+        url.startsWith('mailto:') ||
+        url.startsWith('#')
```
nathanosdev, line 120: I'm not sure why this was moved back. I recently removed it to fix the internal links in the API docs, are they still working with this change?

docs/astro.config.mjs
ilbertt: I just wanted to make it explicit, now that we really need it

docs/src/libs-plugin.ts
ilbertt: The reason is that we have two cases here:
- links that start with `#`, which we want to leave as it is because it's a hash link to the same page
- links that end with an hash link `#...`, which we want to normalize as we still do at line 129 (you fixed it in https://github.com/dfinity/agent-js/pull/1080/files#diff-53024d6ed6e8cd022a250eb37444767d17325704405f05a564bcba5d71725b8eR127)

dfx.json
ilbertt: I thought that local testing was done with `pnpm start` or `pnpm preview` from the docs folder, without the need for a canister.

ilbertt: I'll add instructions on how to do local deployments with Juno

ilbertt: Juno does not have a simple way to deploy a satellite locally, without going through many configuration steps before. Let's move on, and I'll update the contributing guidelines once Juno fixes it.

nathanosdev (approved):

PR #1090, opened 2025-08-04
docs: add dfinity starlight theme

nathanosdev:
# Description

This PR adds a Starlight plugin with a Dfinity-style theme. It's not perfect (I'm not generally good with colors), but it's much closer to the dev docs now than it was before and this gives a starting point for iterating on that design.

Fixes # (issue)

# How Has This Been Tested?

# Checklist:


docs/src/plugins/theme/dfinity-starlight-theme.css
nathanosdev, line 1: This is based on a combination of the colors from the portal and the default starlight colors. I couldn't directly map the docusaurus color scheme so I anyway think it will be better to get someone from design to set the values in this file.

docs/src/plugins/theme/dfinity-starlight-theme.ts
```diff
@@ -0,0 +1,24 @@
...
+export function dfinityStarlightTheme(): StarlightPlugin {
+  return {
+    name: '@dfinity/starlight-theme',
+    hooks: {
+      'config:setup': ctx => {
+        ctx.updateConfig({
+          logo: {
+            src: './src/plugins/theme/dfinity.svg',
+            alt: 'DFINITY Logo',
+          },
```
ilbertt: Since we are publishing the **ICP** SDK JS docs, shouldn't we call it ICP logo instead?
```suggestion
          logo: {
            src: './src/plugins/theme/icp.svg',
            alt: 'Internet Computer Logo',
          },
```

docs/src/plugins/theme/dfinity-starlight-theme.css
ilbertt: > it will be better to get someone from design to set the values in this file

Agree

docs/src/plugins/theme/dfinity-starlight-theme.ts
nathanosdev: Done, thanks!

PR #1092, opened 2025-08-05
fix: error handling of call responses

ilbertt:
# Description

- change error code to `UncertifiedRejectErrorCode` when throwing an error in v2 responses
- fixes retry logic to avoid syncing time indefinitely in case of an ingress expiry error

# How Has This Been Tested?

Added e2e test for ingress expiry fix.

# Checklist:

### DIFF
--- a/packages/agent/src/actor.ts
+++ b/packages/agent/src/actor.ts
@@ -469,20 +469,20 @@ function _createActorMethod(
           }
         }
       } else if (isV2ResponseBody(response.body)) {
-        // handle v2 response errors by throwing an UpdateCallRejectedError object
         const { reject_code, reject_message, error_code } = response.body;
-        const certifiedRejectErrorCode = new CertifiedRejectErrorCode(
+        const errorCode = new UncertifiedRejectErrorCode(
           requestId,
           reject_code,
           reject_message,
           error_code,
+          undefined,
         );
-        certifiedRejectErrorCode.callContext = {
+        errorCode.callContext = {
           canisterId: cid,
           methodName,
           httpDetails: response,
         };
-        throw RejectError.fromCode(certifiedRejectErrorCode);
+        throw RejectError.fromCode(errorCode);
       }
 
       // Fall back to polling if we receive an Accepted response code
--- a/packages/agent/src/agent/http/index.ts
+++ b/packages/agent/src/agent/http/index.ts
@@ -481,6 +481,7 @@ export class HttpAgent implements Agent {
    * @param options.effectiveCanisterId - (Optional) The effective canister ID, if different from the target canister ID.
    * @param options.callSync - (Optional) Whether to use synchronous call mode. Defaults to true.
    * @param options.nonce - (Optional) A unique nonce for the request. If provided, it will override any nonce set by transforms.
+   * @param options.throwOnIngressExpiryError - (Optional) Whether to throw an error if the ingress expiry is invalid. If `false`, retries the request after syncing time with the IC network. Default `false`.
    * @param identity - (Optional) The identity to use for the call. If not provided, the agent's current identity will be used.
    * @returns A promise that resolves to the response of the call, including the request ID and response details.
    */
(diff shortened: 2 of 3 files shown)


packages/agent/src/agent/http/index.ts
```diff
@@ -481,6 +481,7 @@ export class HttpAgent implements Agent {
    * @param options.effectiveCanisterId - (Optional) The effective canister ID, if different from the target canister ID.
    * @param options.callSync - (Optional) Whether to use synchronous call mode. Defaults to true.
    * @param options.nonce - (Optional) A unique nonce for the request. If provided, it will override any nonce set by transforms.
+   * @param options.throwOnIngressExpiryError - (Optional) Whether to throw an error if the ingress expiry is invalid. If `false`, retries the request after syncing time with the IC network. Default `false`.
```
mraszyk: Does it make sense for this to be a public input arg?

packages/agent/src/actor.ts
```diff
@@ -469,20 +469,20 @@ function _createActorMethod(
...
       } else if (isV2ResponseBody(response.body)) {
-        // handle v2 response errors by throwing an UpdateCallRejectedError object
         const { reject_code, reject_message, error_code } = response.body;
-        const certifiedRejectErrorCode = new CertifiedRejectErrorCode(
+        const errorCode = new UncertifiedRejectErrorCode(
           requestId,
           reject_code,
           reject_message,
           error_code,
+          undefined,
```
mraszyk: The fact that we may include signatures in an uncertified reject is a bit confusing.

packages/agent/src/actor.ts
```diff
@@ -520,18 +520,25 @@ function _createActorMethod(
...
-          : undefined;
+        const errorCode = new UnexpectedErrorCode(
+          'Call was returned undefined, and the method has no return type. We cannot determine if the call was successful or not.',
+        );
+        errorCode.callContext = {
+          canisterId: cid,
+          methodName,
+          httpDetails,
+        };
+        throw UnknownError.fromCode(errorCode);
```
mraszyk: Why special case no return type?

packages/agent/src/agent/http/index.ts
ilbertt: We don't have many choices here since the function is exposed publicly

mraszyk: We could introduce a private helper with an extra arg.

packages/agent/src/actor.ts
ilbertt: That's for queries, where we want to include signatures in the error for devs to act upon, if they need

mraszyk: Right, but a separate variant for update calls (that do not provide any signatures) might be less confusing.

packages/agent/src/actor.ts
mraszyk: i.e., I'd just drop this if and fallback in the final else branch

packages/agent/src/agent/http/index.ts
ilbertt: I don't think it's worth the refactor effort. @nathanosdev what do you think?

packages/agent/src/actor.ts
ilbertt: Done

packages/agent/src/actor.ts
ilbertt: Added `UncertifiedRejectUpdateErrorCode`.

packages/agent/src/agent/http/index.ts
ilbertt: @mraszyk how about we check for `hasSyncedTime` instead of passing around the `throwOnIngressExpiryError` flag?

nathanosdev: What about using `internal`? https://tsdoc.org/pages/tags/internal/

mraszyk: > how about we check for hasSyncedTime instead of passing around the throwOnIngressExpiryError flag?

That's a great idea!

ilbertt: Done

nathanosdev (approved):

PR #1093, opened 2025-08-05
fix: avoid bigint overflow when decoding the time from the certificate

ilbertt:
# Description

Removes potential bug with BigInt overflow when decoding the time from a certificate.

# How Has This Been Tested?

Same tests should still pass.

# Checklist:


packages/agent/src/utils/leb.ts
nathanosdev, line 1: A unit test to assert the overflow doesn't happen would be nice

ilbertt: Testing the Number overflow with a Date is tricky. This fix probably doesn't add much, as we would hit a `Date` overflow way before a `Number(bigint)` overflow. Anyway, at least code is cleaner

PR #1102, opened 2025-08-19
feat: `@icp-sdk/core` upgrading guide and migrator

ilbertt:
# Description

Adds the upgrading guide to the docs and a new migrator package to help people migrating to the new package.

Blocked by #1105.

# How Has This Been Tested?

Ran it locally on both a dfx project and [dfinity/ic-js](https://github.com/dfinity/ic-js).

# Checklist:


docs/src/content/docs/upgrading.md
```diff
@@ -0,0 +1,139 @@
+# How to upgrade to @icp-sdk/core
+
+The new `@icp-sdk/core` package is made up of multiple submodules, each of which contains the functionality of the old `@dfinity/...` package.
+
+Tree-shaking is supported, so you can only import the submodules you need.
```
nathanosdev:
```suggestion
Tree-shaking is supported, so only the code you import is included in your final bundle.
```

packages/migrate/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@icp-sdk/migrate",
+  "version": "1.0.0-beta.0",
+  "author": "DFINITY Stiftung <sdk@dfinity.org>",
+  "license": "Apache-2.0",
+  "description": "CLI tool to migrate from agent-js to @icp-sdk/core",
+  "homepage": "https://js.icp.build/core/latest/migrate",
```
nathanosdev: Will it be added in a follow up?

ilbertt: Updated the URL

docs/src/content/docs/upgrading.md
ilbertt: Moved to the FAQ section

packages/migrate/src/code-migrator.ts
```diff
@@ -0,0 +1,114 @@
...
+        !relativePath.startsWith('node_modules') &&
+        !relativePath.startsWith('dist') &&
+        !relativePath.startsWith('build') &&
+        !relativePath.startsWith('.git')
+      );
+    });
+  }
+
+  private async migrateFile(filePath: string): Promise<{ importsReplaced: number }> {
+    if (!existsSync(filePath)) {
```
nathanosdev: https://nodejs.org/docs/latest/api/fs.html#fsexistspath-callback

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

nathanosdev (approved):

packages/migrate/src/code-migrator.ts
ilbertt: Fixed

PR #1108, opened 2025-08-20
fix: add `bigint` to the `JsonValue` types

ilbertt:
# Description

Candid and cbor decoding can return bigints, so we must add that type to the return type.

Fixes https://github.com/demergent-labs/azle/issues/2639.

# How Has This Been Tested?

Same tests should pass.

# Checklist:

### DIFF
--- a/packages/candid/src/types.ts
+++ b/packages/candid/src/types.ts
@@ -4,4 +4,4 @@ export interface JsonArray extends Array<JsonValue> {}
 
 export interface JsonObject extends Record<string, JsonValue> {}
 
-export type JsonValue = boolean | string | number | JsonArray | JsonObject;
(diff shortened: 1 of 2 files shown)


packages/candid/src/types.ts
```diff
@@ -4,4 +4,4 @@ export interface JsonArray extends Array<JsonValue> {}
 
 export interface JsonObject extends Record<string, JsonValue> {}
 
-export type JsonValue = boolean | string | number | JsonArray | JsonObject;
+export type JsonValue = boolean | string | number | bigint | JsonArray | JsonObject;
```
peterpeterparker, line 7: Not sure about the context, but out of curiosity: is this spec-compliant? Or could it cause issues and lead to confusion, since `bigint` is not a valid JSON value?

PR #1152, opened 2025-09-19
fix(identity): handle ArrayBuffer in delegation chain serialization

DrJesseGlass:
# Description

Fixes 'Uint8Array expected' error in `DelegationChain.toJSON()` caused by `@noble/hashes` v1.8+ requiring strict `Uint8Array` inputs.

**Root Cause**: The upgrade to `@noble/hashes` v1.8+ introduced stricter type checking where `bytesToHex()` only accepts `Uint8Array`, but delegation serialization can encounter `ArrayBuffer` objects from:
- Browser crypto APIs returning `ArrayBuffer` instead of `Uint8Array`
- Serialization/deserialization cycles converting between storage formats
- External libraries passing mixed binary data types

**Impact**: Applications using delegation chain persistence (like `ic-siws-js`) crash with "Uint8Array expected" when calling `toJSON()` on delegation chains containing `ArrayBuffer` data.

**Solution**: Added `safeBytesToHex()` wrapper that converts `ArrayBuffer` to `Uint8Array` before calling `bytesToHex()`. This maintains backward compatibility while working with the stricter noble library requirements.

## Changed Methods:
- `Delegation.toJSON()` - handles `ArrayBuffer` in `pubkey` field
- `DelegationChain.toJSON()` - handles `ArrayBuffer` in `signature` and `publicKey` fields

# How Has This Been Tested?
- **All existing tests pass**: 22 identity package tests continue to work
- **Regression test added**: Simulates `ArrayBuffer` conditions that caused the original failure
- **Integration verified**: Resolves the issue in `ic-siws-js` that triggered this investigation
- **Type compatibility**: Handles both `Uint8Array` and `ArrayBuffer` inputs seamlessly
- **Round-trip tested**: `toJSON()` → `fromJSON()` → `toJSON()` cycles work correctly

**Test command**: `pnpm --filter @dfinity/identity test`

**Integration test**: Verified fix resolves the original issue in `ic-siws-js` component that triggered this investigation.

# Checklist:

- [X] My changes follow the guidelines in [CONTRIBUTING.md](https://github.com/dfinity/icp-js-core/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/packages/identity/src/identity/delegation.test.ts
+++ b/packages/identity/src/identity/delegation.test.ts
@@ -28,11 +28,9 @@ expect.extend({
   },
 });
 
-declare global {
-  namespace jest {
-    interface Matchers<R> {
-      toBeHex(): R;
-    }
+declare module '@jest/expect' {
+  interface Matchers<R> {
+    toBeHex(): R;
   }
 }
 


packages/candid/package.json
```diff
@@ -52,5 +52,8 @@
   },
   "peerDependencies": {
     "@dfinity/principal": "workspace:*"
+  },
+  "dependencies": {
+    "@noble/hashes": "^1.8.0"
```
ilbertt: Let's keep it in the `peerDependencies`, like we do for the other submodules/packages.

Afterwards, you must run `pnpm install` again to update the lock file.

ilbertt (changes requested):
@DrJesseGlass thanks for the contribution and the detailed explanation in the PR!

I'm wondering: shouldn't the usage of `ArrayBuffer` instead of `Uint8Array` be caught by the TypeScript checks before runtime? In fact, in the tests that you added, you explicitly cast the type to `any` to avoid TS errors.

I think the proper fix is to convert to `Uint8Array` downstream, like you did in https://github.com/kristoferlund/ic-siws/pull/6, leaving the `@icp-sdk/core/identity` code untouched.

packages/candid/package.json
DrJesseGlass: Thanks for the catch. This was an accidental artifact of when I was exploring @noble/hashes 1.8.0 issues. It should not be added to the candid/package.json at all.

packages/identity/src/identity/delegation.test.ts
```diff
@@ -155,3 +155,104 @@ describe('PartialDelegationIdentity', () => {
...
+      chain.publicKey
+    );
+
+    // This tests another code path that could fail with ArrayBuffer
+    expect(() => {
+      const json = chainWithArrayBufferPubkey.toJSON();
+      expect(json.delegations[0].delegation.pubkey).toMatch(/^[0-9a-f]+$/);
+    }).not.toThrow();
+  });
+});
```
ilbertt: Can you add a newline here?
```suggestion
});

```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- fix(identity): Fix 'Uint8Array expected' error in DelegationChain serialization with ArrayBuffer data
```
ilbertt: How about we use the same title as in the PR?

```suggestion
- fix(identity): handle ArrayBuffer in delegation chain serialization
```

packages/identity/src/identity/delegation.test.ts
```diff
@@ -155,3 +155,104 @@ describe('PartialDelegationIdentity', () => {
     });
   });
 });
+
+describe('DelegationChain ArrayBuffer serialization bug fix', () => {
```
ilbertt:
```suggestion
describe('DelegationChain with ArrayBuffers', () => {
```

packages/identity/src/identity/delegation.test.ts
```diff
@@ -155,3 +155,104 @@ describe('PartialDelegationIdentity', () => {
...
+    );
+
+    const chainWithArrayBuffers = DelegationChain.fromDelegations(
+      delegationsWithArrayBuffer,
+      publicKeyArrayBuffer as ArrayBuffer
+    );
+
+    // This would throw "Uint8Array expected" before the safeBytesToHex fix
+    // but should work fine after the fix
+    expect(() => {
```
ilbertt, line 235:
```suggestion
    expect.assertions(3)
    // This would throw "Uint8Array expected" before the safeBytesToHex fix
    // but should work fine after the fix
    expect(() => {
```

packages/identity/src/identity/delegation.test.ts
```diff
@@ -155,3 +155,104 @@ describe('PartialDelegationIdentity', () => {
...
+      };
+    });
+
+    const chainWithArrayBufferPubkey = DelegationChain.fromDelegations(
+      delegationsWithArrayBufferPubkey,
+      chain.publicKey
+    );
+
+    // This tests another code path that could fail with ArrayBuffer
+    expect(() => {
```
ilbertt, line 281:
```suggestion
    expect.assertions(2);
    // This tests another code path that could fail with ArrayBuffer
    expect(() => {
```

packages/identity/src/identity/delegation.test.ts
```diff
@@ -155,3 +155,104 @@ describe('PartialDelegationIdentity', () => {
...
+      publicKeyArrayBuffer as ArrayBuffer
+    );
+
+    // This would throw "Uint8Array expected" before the safeBytesToHex fix
+    // but should work fine after the fix
+    expect(() => {
+      const json = chainWithArrayBuffers.toJSON();
+      // Verify the output is still valid hex
+      expect(json.delegations[0].signature).toMatch(/^[0-9a-f]+$/);
+      expect(json.publicKey).toMatch(/^[0-9a-f]+$/);
```
ilbertt: Minor thing: let's extract this into a function that we can reuse in both tests. You can place the function at the end of the file:
```ts
function isHex(value: string) {
  if (typeof value !== 'string' || value.length === 0) {
    return false;
  }
  return /^[0-9a-f]+$/i.test(value); // <- Note the "i" regex flag to make it case insensitive
}

// ...

expect(isHex(json.publicKey)).toBe(true);
```

A better pattern would probably be with [`expect.extend`](https://jestjs.io/docs/expect#expectextendmatchers), but let's skip it to avoid adding complexity.

ilbertt (dismissed):
I've thought more about this and I think it makes sense to avoid developers having to manually check for the proper buffer type.

I left a few minor comments that I think can be resolved easily.

packages/identity/src/identity/delegation.test.ts
DrJesseGlass: I implemented the requested changes. But went with expect.extend -- will happily convert to `function isHex(value: string)` with `expect(isHex(json.delegations[0].signature)).toBe(true);` uses. Just figured to lean into what you felt was the better pattern.

ilbertt (approved):

packages/identity/src/identity/delegation.test.ts
ilbertt: Great!

ilbertt (approved):

ilbertt:
@DrJesseGlass merged! Thanks a lot for the contribution and the iterations on the feedback. A patch containing the fix will be released soon.

PR #1156, opened 2025-10-05
feat(agent): `getCanisterEnv` and `safeGetCanisterEnv` (experimental)

ilbertt:
# Description

Introduces the `getCanisterEnv` and `safeGetCanisterEnv` functions to load the canister environment from the `ic_env` cookie. This is an _experimental_ feature.

Additionally, adds the `@types/jsdom` package to the dev dependencies to avoid typescript errors.

Related: https://github.com/dfinity/sdk/pull/4387.

# How Has This Been Tested?

Introduced ad-hoc unit tests.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- feat(agent): introduce the `getCanisterEnv` and `safeGetCanisterEnv` functions to load the canister environment from the `ic_env` cookie.
+- feat(agent): introduce the `getCanisterEnv` and `safeGetCanisterEnv` functions to load the canister environment from the `ic_env` cookie. (experimental)
 
 ## [4.1.1] - 2025-10-21
 


packages/agent/src/agent/http/canister-env.ts
```diff
@@ -0,0 +1,186 @@
...
+   * // The `@icp-sdk/bindgen` package has a feature to automatically generate the .d.ts file
+   * declare global {
+   *   interface CanisterEnv {
+   *     readonly ['PUBLIC_CANISTER_ID:backend']: string;
+   *   }
+   * }
+   *
+   * const env = getCanisterEnv();
+   *
+   * console.log(env.IC_ROOT_KEY); // by default served by the canister
```
viviveevee: Is `the canister` understandable in this context? Out of context it sounds like every canister ever would serve the cookie, but only the asset canister does that

packages/agent/src/agent/http/canister-env.ts
```diff
@@ -0,0 +1,186 @@
...
+ * const env = safeGetCanisterEnv();
+ * console.log(env); // undefined
+ * ```
+ */
+export function safeGetCanisterEnv<T = Record<string, never>>(
+  options: GetCanisterEnvOptions = {},
+): (CanisterEnv & T) | undefined {
+  if (!globalThis.document) {
+    return undefined;
+  }
```
viviveevee: Is this check really expensive enough to justify having separate `getCanisterEnv` and `safeGetCanisterEnv`?

packages/agent/src/agent/http/canister-env.test.ts
```diff
@@ -0,0 +1,314 @@
...
+        expect(bytesToHex(env.IC_ROOT_KEY)).toBe(mockRootKeyHex);
+      });
+
+      test('should properly decode URI-encoded cookie values', () => {
+        type TestCanisterEnv = {
+          readonly IC_ROOT_KEY: Uint8Array;
+          readonly PUBLIC_MESSAGE: string;
+        };
+
+        const specialChars = 'Hello World!@#$%';
```
viviveevee:
```suggestion
        const specialChars = 'Hello World!=@#$%';
```
You are explicitly only searching for the first `=` as the separator, right? So we probably should try throwing in an `=` here too

ilbertt: Done

packages/agent/src/agent/http/canister-env.ts
ilbertt: I've updated the doc comments. Feel free to resolve this comment if that's enough

packages/agent/src/agent/http/canister-env.ts
ilbertt: I think I should refactor `getCanisterEnv` to always return a canister env and throw if something is wrong. This way, calling `getCanisterEnv` in a frontend application doesn't require the dev to check if the result is `undefined`, which should never be in a browser environment when assets are served by an asset canister that supports this feature.

ilbertt: I've refactored `getCanisterEnv`, feel free to resolve this comment if the refactor is enough

sesi200 (approved):

docs/src/content/docs/_sidebar.json
```diff
@@ -4,7 +4,16 @@
...
+  },
+  {
+    "label": "Guides",
+    "items": [
+      {
+        "label": "Canister Environment",
+        "link": "/canister-environment"
+      },
       { "label": "Typescript", "link": "/typescript" },
       { "label": "Library Development", "link": "/library-development" }
```
ilbertt, line 22: Moves the "Typescript" and "Library Development" to the new "Guides" section

viviveevee (approved):

PR #1345, opened 2026-04-10
feat: add optional `onPollingStarted` callback when making update calls

hpeebles:
# Description

This allows devs to specify an optional callback when submitting update calls, which will be invoked once their request has been accepted by the IC.

We use this within OpenChat to display the first message tick, signalling that the message has made its way to the backend.

# How Has This Been Tested?

I have added tests which cover the various cases where the callback should either be invoked or not, and also we are using this successfully within OpenChat.

# Checklist:

### DIFF
--- a/packages/core/src/agent/agent/api.ts
+++ b/packages/core/src/agent/agent/api.ts
@@ -130,14 +130,15 @@ export interface CallOptions {
 
 export interface UpdateOptions extends CallOptions {
   /**
-   * An optional callback that will be invoked once the request is accepted by the IC.
+   * An optional callback that will be invoked once the agent starts polling for the result of the
+   * update call.
    *
-   * If `callSync` is set to false, the callback will be invoked once the request has been accepted
-   * by a single node. If `callSync` is set to true, the IC will first try to process the request
-   * synchronously, but if the synchronous request timeout is exceeded, the IC will reply
-   * indicating the request was accepted, at which point the callback will be invoked.
+   * If `callSync` is set to false, polling will start, and the callback will be invoked, once the
+   * request has been accepted by a single node. If `callSync` is set to true, the IC will first
+   * try to process the request synchronously. But if the synchronous request timeout is exceeded,
+   * the agent will start polling for the result, at which point the callback will be invoked.
    */
-  onRequestAccepted?: () => void;
+  onPollingStarted?: () => void;
 }
 
 export interface ReadStateResponse {
--- a/packages/core/src/agent/agent/api.ts
+++ b/packages/core/src/agent/agent/api.ts
@@ -118,6 +118,11 @@ export interface CallOptions {
   effectiveCanisterId: Principal | string;
 
   /**
+   * Whether to use synchronous call mode. Defaults to true.
+   */
+  callSync?: boolean;
+
+  /**
    * An optional nonce to use for the call, used to prevent replay attacks.
    */
   nonce?: Uint8Array;
--- a/packages/core/src/agent/agent/api.ts
+++ b/packages/core/src/agent/agent/api.ts
@@ -123,6 +123,14 @@ export interface CallOptions {
   nonce?: Uint8Array;
 }
 
+export interface UpdateOptions extends CallOptions {
+  /**
+   * An optional callback that will be invoked once the request is accepted by the IC.
+   * Using this requires `callSync` to be set to false.
+   */
+  onRequestAccepted?: () => void;
+}
+
 export interface ReadStateResponse {
   certificate: Uint8Array;
 }
--- a/packages/core/src/agent/agent/api.ts
+++ b/packages/core/src/agent/agent/api.ts
@@ -131,7 +131,11 @@ export interface CallOptions {
 export interface UpdateOptions extends CallOptions {
   /**
    * An optional callback that will be invoked once the request is accepted by the IC.
-   * Using this requires `callSync` to be set to false.
+   *
+   * If `callSync` is set to false, the callback will be invoked once the request has been accepted
+   * by a single node. If `callSync` is set to true, the IC will first try to process the request
+   * synchronously, but if the synchronous request timeout is exceeded, the IC will reply
+   * indicating the request was accepted, at which point the callback will be invoked.
    */
   onRequestAccepted?: () => void;
 }
--- a/packages/core/src/agent/agent/http/http-update.test.ts
+++ b/packages/core/src/agent/agent/http/http-update.test.ts
@@ -445,10 +445,10 @@ describe('HttpAgent.update', () => {
     });
   });
 
-  it('invokes onRequestAccepted callback', async () => {
+  it('invokes onPollingStarted callback', async () => {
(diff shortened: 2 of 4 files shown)


packages/core/src/agent/agent/api.ts
```diff
@@ -117,12 +117,25 @@ export interface CallOptions {
...
   /**
    * An optional nonce to use for the call, used to prevent replay attacks.
    */
   nonce?: Uint8Array;
 }
 
+export interface UpdateOptions extends CallOptions {
+  /**
+   * An optional callback that will be invoked once the request is accepted by the IC.
+   * Using this requires `callSync` to be set to false.
```
peterpeterparker: None of my business, I randomly spotted this PR, but, have you considered using a variant for introducing this new option?

With the current pattern, you'll need to implement an assertion for this requirement (it seems there is none yet), and developers would have to read the documentation to understand the usage. Both could likely be resolved with a variant.

hpeebles: Do you mean replacing `callSync` and `onRequestAccepted` with a single variant?
If so, I guess it is possible, but `callSync` already exists and is a simple boolean.
So switching to a variant would be a breaking change, but also I think it would likely make the args harder to understand than the separate fields which are individually each very easy to understand.

hpeebles: An assertion probably isn't needed here.
If you set `callSync: false`, then the callback will be triggered at the speed of a query request, if you set it to true then the request will resort to polling after some duration (10 seconds?) at which point the callback will be invoked.
So in theory, someone may want to use the callback with `callSync` set to true so that they can detect when their request has resorted to polling.

peterpeterparker: > Do you mean replacing callSync and onRequestAccepted with a single variant?

For example yes. Though it does not necessary have to be a breaking change, `callSync` for example can remain a `boolean` or a variant (`typeof === "boolean'`).

Anyway, just my two cents, was sharing if interesting if it hadn't been considered. I'll close this thread then.

hpeebles: I've updated my comment to detail the different behaviours based on the value of `callSync`.

hpeebles: > > Do you mean replacing callSync and onRequestAccepted with a single variant?
> 
> For example yes. Though it does not necessary have to be a breaking change, `callSync` for example can remain a `boolean` or a variant (`typeof === "boolean'`).
> 
> Anyway, just my two cents, was sharing if interesting if it hadn't been considered. I'll close this thread then.

Ah yes I see, I guess that variant trick would work well if we definitely wanted to restrict it to only being set when `callSync` is true.

peterpeterparker: Sounds good. Thx for the feedback.

packages/core/src/agent/agent/http/index.ts
```diff
@@ -681,6 +671,7 @@ export class HttpAgent implements Agent {
     }
 
     if (response.status === HTTP_STATUS_ACCEPTED) {
+      fields.onRequestAccepted?.();
```
raymondk, line 674: is this consistent with the comment?
I don't understand how a timeout gets you here and it seems to be the only place where the callback used.

Maybe a little comment with an explanation would help future generations?

hpeebles: If `callSync` is true, the request to `this.call(..)` will wait for some duration (10 seconds?) for the call to complete, if it completes in that time then we get back a V4 response, if not we get back a HTTP_STATUS_ACCEPTED status code and we must then poll for the result.
If `callSync` is false, the synchronous waiting is skipped and we go straight to polling.

raymondk: What about hitting `case: undefined` inside https://github.com/hpeebles/agent-js/blob/e1e7f9e20eda282154fd2c7a1d4903fd2d8bbffe/packages/core/src/agent/agent/http/index.ts#L657

Doesn't that fall back on polling and then the callback never fires or is there something that prevents that particular combination from happening?

nikosxenakis: With that said, it may be worth considering covering both cases and renaming the callback to better reflect what's happening. In the v4 `case: undefined` path, we don't yet know if a rejection will follow, so "accepted" is not guaranteed. Renaming to something like `onPollingStarted` or `onAwaitingResult` would cover both the 202 and v4 fallback paths and make the contract clearer to callers. This essentially fires when we're about to poll. Also, additional test cases covering rejection paths are required. Thoughts?

hpeebles: @raymondk right yes good point, although that case shouldn't actually be possible according to the IC spec.

See here - https://docs.internetcomputer.org/references/ic-interface-spec#http-sync-call-overview

"If an implementation specific timeout for the request is reached while the replica waits for the terminal state, then the replica will reply with an empty body and a 202 HTTP status code."

And also here - https://docs.internetcomputer.org/references/ic-interface-spec#http-call

"200 HTTP status with a non-empty body. This status is returned if the canister call completed within an implementation-specific timeout or was rejected within an implementation-specific timeout."

So according to those sections of the spec, a V4 response should either contain the actual response or an error, so the `undefined` case is not a valid scenario.

If we simply delete the `undefined` arm of the switch statement, then the `onRequestAccepted` logic becomes valid and covers all paths where the request has been accepted by the IC.

nikosxenakis: According to the spec you are absolutely right. Recently though we had cases where IC was not behaving as expected. Resulting to certificates that didn't contain a response for the requestId, that the reason we recently added polling in this case and we should handle it regardless.

hpeebles: Done. (Finally!)

packages/core/src/agent/agent/http/http-update.test.ts
```diff
@@ -444,4 +444,11 @@ describe('HttpAgent.update', () => {
       expect(agent.readState).not.toHaveBeenCalled();
     });
   });
+
+  it('invokes onPollingStarted callback', async () => {
+    const agent = createAgentWithCallMock();
+    let callbackInvoked = false;
+    await agent.update(canisterId, { ...updateFields, callSync: false, onPollingStarted: () => callbackInvoked = true });
+    expect(callbackInvoked).toBe(true);
+  });
```
nikosxenakis: Could we also add dedicated tests to cover
1. v4 fallback path (path 3) where the certificate doesn't contain the request ID
2. callSync: true with successful sync response, where the callback should not fire
3. Rejection scenario, verifying callback behavior when the call is rejected

hpeebles: Done

nikosxenakis (dismissed):
Looks good, thank you.

nikosxenakis:
@hpeebles please resolve the conflict and I can merge the changes.

nikosxenakis (approved):

PR #1356, opened 2026-04-29
fix(identity): skip sender_info injection for read_state requests

### LINKED ISSUE #1355: AttributesIdentity injects sender_info into read_state polls, breaking ingress sig verification on every update call
## Summary

[`AttributesIdentity.transformRequest`](https://github.com/dfinity/icp-js-core/blob/main/packages/core/src/identity/identity/attributes.ts) unconditionally injects `sender_info` into the body of every outgoing `HttpAgentRequest`. That includes the `read_state` polls the agent fires after submitting an update call. The IC's `read_state` hash does **not** include `sender_info` in its representation-independent hash, so the polled request's basic signature ends up over `hash(body+sender_info)` while the IC verifies against `hash(body)` — every poll fails with:

```
Status: 400
Body: Invalid signature: Invalid basic signature: Ed25519 signature could not be verified: …
```

Net effect: any update call made through `AttributesIdentity` succeeds at `/api/v4/.../call` but then the agent can never poll the response. To the caller it looks like every `actor.method()` rejects.

This affects every consumer of `AttributesIdentity` that calls a non-query canister method, on `@icp-sdk/core@^5.x` (verified against `5.3.1`).

## Reproduction

```ts
import { HttpAgent, Actor } from '@icp-sdk/core/agent';
import { AuthClient } from '@icp-sdk/auth/client';
import { AttributesIdentity, Ed25519KeyIdentity } from '@icp-sdk/core/identity';
import { Principal } from '@icp-sdk/core/principal';

const session = Ed25519KeyIdentity.generate();
const auth = new AuthClient({
  identity: session,
  identityProvider: 'https://id.ai/authorize',
});

const [inner, attrs] = await Promise.all([
  auth.signIn(),
  pendingNonce.then((nonce) =>
    auth.requestAttributes({ keys: ['sso:dfinity.org:email'], nonce }),
  ),
]);

const identity = new AttributesIdentity({
  inner,
  attributes: attrs,
  signer: { canisterId: Principal.fromText('rdmx6-jaaaa-aaaaa-aaadq-cai') }, // II prod
});
const agent  = await HttpAgent.create({ host: 'https://icp0.io', identity });
const actor  = Actor.createActor(idl, { agent, canisterId: someCanister });

await actor.someUpdate(); // 400 Invalid basic signature on the read_state poll
```

The bare-delegation flow (skip `AttributesIdentity`, use `inner` directly) works fine.

## Wire-level evidence

After instrumenting `globalThis.fetch` and dumping the CBOR-encoded request bodies for one `actor.register()` call I see exactly four request bodies posted:

- 1 × `/api/v4/<canister>/call`     (4595 bytes) — both \`sender_info\` and a valid sender_sig
- 3 × `/api/v2/<canister>/read_state` (4580 bytes each) — also has \`sender_info\` in `content`, plus a different sender_sig

The IC's 400 error reports the **read_state's** sender_sig (matches the sig field bytes in the read_state body, NOT the /call). I.e. the call itself was accepted; only the poll fails.

Decoding the read_state body's `content` shows:
```
content: a5
  6c 'request_type'    : 6a 'read_state'
  65 'paths'            : 81 [ 82 'request_status' <hash> ]
  66 'sender'           : 581d <29-byte principal>
  6e 'ingress_expiry'   : 1b <u64>
  6b 'sender_info'      : a3 { signer, info, sig }   ← injected by AttributesIdentity
```

## Root cause

[`representation_independent_hash_call_or_query`](https://github.com/dfinity/ic/blob/52cf4d105ed7ae882e45d09d364c961d2fb442d3/rs/types/types/src/messages/http.rs) **includes** `sender_info` in the hashed map.

[`representation_independent_hash_read_state`](https://github.com/dfinity/ic/blob/52cf4d105ed7ae882e45d09d364c961d2fb442d3/rs/types/types/src/messages/http.rs) hashes only `{ request_type, ingress_expiry, paths, sender, nonce? }` — `sender_info` is not part of the read_state schema and is **not** hashed.

But agent-js's `AttributesIdentity.transformRequest`:

```ts
transformRequest(request: HttpAgentRequest): Promise<unknown> {
  return this.#inner.transformRequest({
    ...request,
    body: {
      ...request.body,
      sender_info: { signer, info, sig },   // unconditional
    },
  } as HttpAgentRequest);
}
```

…doesn't look at `request.body.request_type`, so it pollutes the read_state body with `sender_info`. The DelegationIdentity it wraps then computes `requestIdOf(body)` — which now includes the injected field — and signs that hash. The IC re-derives its own hash without the field, and the signatures disagree.

## Suggested fix

Gate the injection on the request type:

```ts
transformRequest(request: HttpAgentRequest): Promise<unknown> {
  const body = request.body as { request_type?: string } | undefined;
  if (body?.request_type !== 'call' && body?.request_type !== 'query') {
    // read_state, etc. — IC doesn't hash sender_info on these endpoints
    return this.#inner.transformRequest(request);
  }
  return this.#inner.transformRequest({
    ...request,
    body: {
      ...request.body,
      sender_info: {
        signer: this.#signer.canisterId.toUint8Array(),
        info: this.#attributes.data,
        sig: this.#attributes.signature,
      },
    },
  } as HttpAgentRequest);
}
```

(Also worth adding a unit test that constructs a fake `HttpAgentRequest` with `request_type: 'read_state'` and asserts the transformed body has no `sender_info`.)

## Workaround

Subclass / replace `AttributesIdentity` locally with a version that does the request-type gate. Verified to fix the bug end-to-end against prod II + a Motoko canister using \`mo:identity-attributes\`'s \`#Authorization\` policy. Reference implementation:

https://github.com/aterga/identity-attributes/blob/feat/use-icp-sdk-auth/demos/bagel/frontend/src/main.ts (search for `CallOnlyAttributesIdentity`)

## Environment

- `@icp-sdk/core@5.3.1`
- `@icp-sdk/auth@6.2.2` (also reproduces with 6.2.1)
- IC replica revision \`52cf4d105ed7ae882e45d09d364c961d2fb442d3\` (subnet `io67a-…-gqe`, deployed 2026-04-26)
- Browsers: Chrome 122 on macOS

sea-snake:
> **Depends on #1357** — rebase onto main after that PR merges.

# Description

Fixes #1355.

Two related bugs caused `AttributesIdentity` to break update calls end-to-end. Both are addressed here:

### 1. `sender_info` was injected into `read_state` polls

`AttributesIdentity.transformRequest` unconditionally injects `sender_info` into every outgoing body. The IC's representation-independent hash for `read_state` does **not** include `sender_info`, so the sender signature ends up over `hash(body + sender_info)` while the IC verifies against `hash(body)` — every poll fails with `Invalid basic signature`.

**Fix:** gate `sender_info` injection on the request endpoint. Only `call` and `query` get the injection (the IC includes `sender_info` in their representation-independent hash); `read_state` is passed through untouched.

### 2. `HttpAgent.call` / `HttpAgent.query` computed `requestId` from the pre-transform body

`HttpAgent.call` (and `.query`) computed the polling/lookup `requestId` from the pre-transform `submit` / `request` object. The IC, however, stores results keyed by the post-transform hash (which includes any fields an Identity decorator injected during `transformRequest`). With `AttributesIdentity` wrapped around an inner identity, the agent polled under the pre-transform hash and the IC stored the result under the post-transform hash — polling never resolved.

**Fix:** compute `requestId` from `transformedRequest.body.content` after the identity transform so it matches the key the IC will use.

# How Has This Been Tested?

- New unit tests in `packages/core/src/identity/identity/attributes.test.ts`:
  - `read_state` request transformed by `AttributesIdentity` does **not** contain `sender_info` in the content.
  - The request ID hash of the decorated `read_state` content equals the hash of the original (un-decorated) body.
- New regression test in `packages/core/src/agent/agent/http/http.test.ts`:
  - With an Identity decorator that injects an extra field into the body, the `requestId` returned by `HttpAgent.call` equals `requestIdOf` of the post-transform `body.content` that gets serialized and signed.

All existing tests continue to pass (`pnpm --filter @icp-sdk/core test` — 630 passed).

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "tsx": "^4.21.0",
     "typedoc": "^0.28.18",
     "typescript": "5.9.3",
-    "vite": "^8.0.8",
+    "vite": "^8.0.9",
     "vitest": "^4.1.3"
   },
   "packageManager": "pnpm@10.30.3+sha512.c961d1e0a2d8e354ecaa5166b822516668b7f44cb5bd95122d590dd81922f606f5473b6d23ec4a5be05e7fcd18e8488d47d978bbe981872f1145d06e9a740017",
--- a/packages/core/src/agent/agent/http/http.test.ts
+++ b/packages/core/src/agent/agent/http/http.test.ts
@@ -118,6 +118,53 @@ test('call', async () => {
   expect(callParams.headers['Content-Type']).toEqual('application/cbor');
 });
 
+test('call requestId matches the post-transform body content hash', async () => {
+  // Regression test for #1355: when an Identity decorator injects extra fields
+  // into the request body (e.g. AttributesIdentity adds `sender_info`), the
+  // returned requestId must hash the post-transform content so it matches the
+  // key the IC uses when storing the call result.
+  const mockFetch: jest.Mock = jest.fn(() =>
+    Promise.resolve(new Response(null, { status: 200 })),
+  );
+
+  const baseIdentity = createIdentity(0);
+  const extraField = new Uint8Array([42, 43, 44]);
+
+  // A minimal Identity decorator that injects an extra field into the body
+  // before the inner identity computes requestIdOf and signs.
+  const decorated: SignIdentity = {
+    getPrincipal: () => baseIdentity.getPrincipal(),
+    getPublicKey: () => baseIdentity.getPublicKey(),
+    sign: (blob: Uint8Array) => baseIdentity.sign(blob),
+    transformRequest: (req: any) =>
+      baseIdentity.transformRequest({
+        ...req,
+        body: { ...req.body, sender_info: extraField },
+      }),
+  } as SignIdentity;
+
+  const canisterId = Principal.fromText('2chl6-4hpzw-vqaaa-aaaaa-c');
+  const httpAgent = new HttpAgent({
+    fetch: mockFetch,
+    host: 'http://127.0.0.1',
+    identity: decorated,
+  });
+
+  const { requestId } = await httpAgent.call(canisterId, {
+    methodName: 'greet',
+    arg: new Uint8Array([]),
+  });
+
+  const callParams = mockFetch.mock.calls[0][1];
+  const requestBody = cbor.decode<Envelope<CallRequest>>(callParams.body);
+
+  // The returned requestId must equal the hash of the body content that the IC
+  // will see and key results under.
+  expect(bytesToHex(requestId)).toEqual(bytesToHex(requestIdOf(requestBody.content)));
+  // Sanity: the injected field is present in the signed content.
+  expect((requestBody.content as any).sender_info).toBeDefined();
+});
+
 test.todo('query');
 
 test('queries with the same content should have the same signature', async () => {
--- a/packages/core/src/agent/agent/http/index.ts
+++ b/packages/core/src/agent/agent/http/index.ts
@@ -555,9 +555,11 @@ export class HttpAgent implements Agent {
     // Apply transform for identity.
     transformedRequest = (await id.transformRequest(transformedRequest)) as HttpAgentSubmitRequest;
 
+    // Compute the request ID from the post-transform content so it matches the
+    // hash the IC computes (e.g. when an Identity injects `sender_info`).
+    const requestId = requestIdOf(transformedRequest.body.content ?? submit);
     const body = cbor.encode(transformedRequest.body) as Uint8Array<ArrayBuffer>;
     const backoff = this.#backoffStrategy();
-    const requestId = requestIdOf(submit);
     try {
       // Attempt v4 sync call
       const requestSync = () => {
@@ -1047,8 +1049,6 @@ export class HttpAgent implements Agent {
       ingress_expiry: ingressExpiry,
     };
 
-    const requestId = requestIdOf(request);
-
     transformedRequest = await this._transform({
       request: {
         method: 'POST',
@@ -1064,6 +1064,9 @@ export class HttpAgent implements Agent {
     // Apply transform for identity.
     transformedRequest = (await id.transformRequest(transformedRequest)) as HttpAgentRequest;
 
+    // Compute the request ID from the post-transform content so it matches the
+    // hash the IC computes (e.g. when an Identity injects `sender_info`).
+    const requestId = requestIdOf(transformedRequest.body.content ?? request);
     const body = cbor.encode(transformedRequest.body) as Uint8Array<ArrayBuffer>;
 
     const args = {
--- a/packages/core/src/identity/identity/attributes.test.ts
+++ b/packages/core/src/identity/identity/attributes.test.ts
@@ -3,7 +3,12 @@ import { requestIdOf } from '#agent';
 import { DelegationChain, DelegationIdentity } from './delegation.ts';
 import { Ed25519KeyIdentity } from './ed25519.ts';
 import { AttributesIdentity } from './attributes.ts';
-import { Endpoint, type HttpAgentSubmitRequest } from '../../agent/agent/http/types.ts';
+import {
+  Endpoint,
+  ReadRequestType,
+  type HttpAgentReadStateRequest,
+  type HttpAgentSubmitRequest,
+} from '../../agent/agent/http/types.ts';
 import { Expiry } from '../../agent/agent/http/expiry.ts';
 
 function createIdentity(seed: number): Ed25519KeyIdentity {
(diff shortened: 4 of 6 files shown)


aterga:
Hit a second related bug while testing this PR end-to-end against prod II + a real Motoko canister using `mo:identity-attributes`'s `#Authorization` policy. Filing here since it's adjacent and lives in `HttpAgent.call`, but happy to split into its own issue if preferred.

## Symptom

After applying this PR locally (verified that `read_state` polls no longer carry `sender_info`), an `actor.someUpdate()` call still hangs. The wire trace shows one `/api/v4/.../call` followed by N `/api/v2/.../read_state` polls; the polls succeed (HTTP 2xx), but they're polling for a `request_id` the IC never stored, so the agent loops on `processing` until its retry budget runs out — looks like infinite loading to the user.

## Verification

I captured the full CBOR-encoded `/call` body off the wire and verified offline with `cbor2` + `pynacl`:

```python
content = env['content']
# IC-spec representation_independent_hash
hash_with_sender_info     = hash_map(content)                                  # a6e0061a…
hash_without_sender_info  = hash_map({k:v for k,v in content.items() if k != 'sender_info'})  # 5ad29a61…
nacl.signing.VerifyKey(raw_pubkey).verify(b'\x0aic-request' + hash_with_sender_info, sender_sig)
# ✓ signature verifies — IC accepts /call, stores result under a6e0061a…
```

The `read_state` body's `paths` references `5ad29a613cb799a267622f4ff4ce8455af2717808f741d9ceca8dc4dabc8a1f4` — i.e. the hash *without* `sender_info`.

The IC stores the result under `a6e0061a…` (with sender_info). The agent polls for `5ad29a61…` (without). Mismatch → polling never finds the result → infinite retries.

## Root cause

In `packages/core/src/agent/agent/http/index.ts`, `HttpAgent.call` (around the line that builds the v4 sync request, ~570 in `5.3.x`):

```ts
const submit: CallRequest = {
  request_type: SubmitRequestType.Call,
  canister_id, method_name, arg, sender, ingress_expiry,
};
let transformedRequest = (await this._transform({ ..., body: submit })) as HttpAgentSubmitRequest;
// nonce handling …
transformedRequest = (await id.transformRequest(transformedRequest)) as HttpAgentSubmitRequest;
//                   ^^^^^^^^^^^^^^^^^^^^ AttributeIdentity injects sender_info into the body here
const body      = cbor.encode(transformedRequest.body) as Uint8Array<ArrayBuffer>;
const requestId = requestIdOf(submit);   // ← BUG: computed from PRE-transform body
```

`requestId` ends up over the canonical body without `sender_info`, while `transformedRequest.body.content` (what's actually serialized and signed) is over body + `sender_info`. The IC's `representation_independent_hash` for the call includes `sender_info` (`representation_independent_hash_call_or_query` in [`rs/types/types/src/messages/http.rs`](https://github.com/dfinity/ic/blob/52cf4d105ed7ae882e45d09d364c961d2fb442d3/rs/types/types/src/messages/http.rs)), so it stores the result under the post-transform hash. The agent then polls for the pre-transform hash and never finds it.

The same divergence likely exists for the query path (`HttpAgent.query`/`#callV3`/wherever a polling key is derived from a body that an Identity decorator has since augmented).

## Suggested fix

Compute the polling/lookup key from the post-transform content:

```ts
transformedRequest = (await id.transformRequest(transformedRequest)) as HttpAgentSubmitRequest;
const body = cbor.encode(transformedRequest.body) as Uint8Array<ArrayBuffer>;
// Use the same hash the IC will use to key the result.
const requestId = requestIdOf(
  (transformedRequest.body as { content: Record<string, unknown> }).content,
);
```

(Or, if there's a reason to keep `submit` as the source of truth: declare it `let`, have the Identity contract allow mutating it in place, and document that.)

Worth a regression test that submits a call through `AttributeIdentity` and asserts `requestId` matches the hash of `transformedRequest.body.content`.

## Workaround (current)

For anyone hitting this before the fix lands, our local `AttributeIdentity.transformRequest` mutates `request.body` in place instead of returning a new body. That way the outer `submit` reference inside `HttpAgent.call` sees `sender_info` too, and `requestIdOf(submit)` ends up over the same bytes the IC sees. Reference: https://github.com/aterga/identity-attributes/blob/feat/use-icp-sdk-auth/demos/bagel/frontend/src/main.ts (look for `AttributeIdentity` — note the `(request.body as Record<string, unknown>).sender_info = …` mutation).

End-to-end this fully unblocks the flow: `register()` lands in <1s, the canister returns its result, the agent's polling completes cleanly. Without the workaround, the read_state-injection fix alone isn't enough to make a single `AttributeIdentity`-wrapped update call return.

## Environment

- `@icp-sdk/core@5.3.1`
- `@icp-sdk/auth@6.2.2`
- IC replica revision `52cf4d105ed7ae882e45d09d364c961d2fb442d3` (subnet `io67a-…-gqe`)

aterga (dismissed):

lwshang:
The rename rationale doesn't quite land for me. `sender_info` being singular isn't relevant to the class name.

The more relevant thing is that `SignedAttributes.data` is intentionally opaque bytes that can encode any number of caller claims ("attributes"). The plural in `AttributesIdentity` reflects that accurately.

I'd suggest keeping `AttributesIdentity`. Happy to be convinced otherwise if there's a stronger reason I'm missing.

aterga (dismissed):

sea-snake:
@lwshang Reverted the rename in this PR.

marc0olo (approved):

PR #1360, opened 2026-05-20
feat!: effective subnet ID

adamspofford-dfinity:
This PR implements the new effective subnet ID API, where management canister calls can be targeted at a subnet. The declared API for it is a breaking change: several things that previously assumed a canister ID now take a union type, and I took the opportunity to unify many things that had separate canister/subnet functions into single functions with that same union. In an attempt to maintain 'binary compatibility' all the callsites still check for the old parameter types.

Some other changes were made while I had the opportunity. Specifically, TS 5.9 Uint8Array generics that we had been putting off are now implemented, with everything exactingly specified to return an ArrayBuffer, and make owned copies of potential SharedArrayBuffers. I also went around fixing all the 5.9 deprecation errors, and it then seemed like a good time to refactor the tsconfig setup so that tests were included in the default config instead of the default config being for building.


packages/core/src/agent/agent/http/index.ts
```diff
@@ -1236,25 +1269,29 @@ export class HttpAgent implements Agent {
...
-        const request_status = new TextEncoder().encode('request_status');
-        if (uint8Equals(pathName, request_status)) {
-          return value as RequestId;
-        }
-      }
+    if (typeof effectiveTarget === 'string' || effectiveTarget instanceof Principal) {
+      // compatibility with v5
+      effectiveTarget = { canisterId: Principal.from(effectiveTarget) };
     }
+    await this.#rootKeyGuard();
```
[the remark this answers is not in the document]
adamspofford-dfinity: False.

packages/core/src/agent/utils/target.ts
```diff
@@ -0,0 +1,27 @@
+import { Principal } from '#principal';
+import type { InputTargetPrincipal } from '../agent/api.ts';
+import type { TargetPrincipal } from '../certificate.ts';
+import type { ExpirableStore } from './expirableStore.ts';
+
```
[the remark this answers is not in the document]
adamspofford-dfinity: False.

packages/core/src/agent/utils/bls.ts
```diff
@@ -11,8 +11,9 @@ export let verify: (pk: Uint8Array, sig: Uint8Array, msg: Uint8Array) => boolean
...
  */
 export function blsVerify(pk: Uint8Array, sig: Uint8Array, msg: Uint8Array): boolean {
-  const primaryKey = typeof pk === 'string' ? pk : bytesToHex(pk);
-  const signature = typeof sig === 'string' ? sig : bytesToHex(sig);
-  const message = typeof msg === 'string' ? msg : bytesToHex(msg);
-  return bls12_381.verifyShortSignature(signature, message, primaryKey);
+  const primaryKey = typeof pk === 'string' ? hexToBytes(pk) : pk;
+  const signature = typeof sig === 'string' ? hexToBytes(sig) : sig;
+  const message = typeof msg === 'string' ? hexToBytes(msg) : msg;
+  const blss = bls12_381.shortSignatures;
```
sea-snake, line 17: Nit: Is `blss` a typo?

packages/core/src/agent/cbor.ts
```diff
@@ -22,7 +22,7 @@ function hasCborValueMethod(value: unknown): value is ToCborValue {
  * the {@link ToCborValue.toCborValue} method will be called to get the value to encode.
  * @param value The value to encode
  */
-export function encode(value: unknown): Uint8Array {
+export function encode(value: unknown): Uint8Array<ArrayBuffer> {
```
sea-snake: Is there a reason to expose the underlying buffer in the type? Ideally, such implementation details are omitted for simplicity and to avoid breaking future clients when the underlying buffer needs to change.

packages/core/package.json
```diff
@@ -93,14 +101,16 @@
...
+    "@noble/curves": "^2.2.0",
+    "@noble/hashes": "^2.2.0",
+    "@scure/bip32": "^2.2.0",
+    "@scure/bip39": "^2.2.0",
     "asn1js": "^3.0.7"
   },
   "devDependencies": {
+    "@types/jsdom": "^28.0.0",
     "fake-indexeddb": "^6.2.5",
+    "jsdom": "^28.0.0",
```
sea-snake, line 113: Why was jsdom added?

adamspofford-dfinity: It is already used by tests. I do not know why it previously worked.

packages/core/src/agent/utils/bls.ts
adamspofford-dfinity: This is the name used for it in noble-curves' readme 🤷

sea-snake: Ah it's probably shorthand for BLS signatures I suppose, it's a non exported variable so it's fine.

packages/core/src/agent/cbor.ts
adamspofford-dfinity: This particular function is annotated primarily because the tests require a ton of casting otherwise. Other functions are because we can, and callers will want it that way for other functions with the same requirements.

packages/core/tsconfig.json
```diff
@@ -1,17 +1,15 @@
...
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "composite": true,
-    "baseUrl": "./src",
+    "noEmit": true,
     "rootDir": "./src",
-    "outDir": "./lib/esm",
+    "types": ["jest", "node"],
     "paths": {
```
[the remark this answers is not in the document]
adamspofford-dfinity: This has been false for six years.

packages/core/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "rootDir": "./src",
+    "outDir": "./lib/esm",
+    "paths": {
```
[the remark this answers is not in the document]
adamspofford-dfinity: This has been false for six years.

packages/core/src/candid/utils/buffer.ts
```diff
@@ -177,8 +180,9 @@ export function uint8FromBufLike(
   if ('buffer' in bufLike) {
     return uint8FromBufLike(bufLike.buffer);
   }
-  return new Uint8Array(bufLike);
+  return new Uint8Array(bufLike).slice();
 }
+export type Uint8ArrayBuffer = ReturnType<typeof Uint8Array.of>;
```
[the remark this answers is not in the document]
adamspofford-dfinity: This is false. `Uint8ArrayBuffer` is declared as `Uint8Array<ArrayBuffer>` on TS5.9+, and on older versions, `Uint8Array` which cannot be parameterized.

packages/core/src/agent/utils/bls.ts
```diff
@@ -11,8 +11,9 @@ export let verify: (pk: Uint8Array, sig: Uint8Array, msg: Uint8Array) => boolean
...
+  return blss.verify(signature, blss.hash(message), primaryKey);
```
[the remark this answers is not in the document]
adamspofford-dfinity: Backwards compatibility code for types incompatible with declared types is explicitly intended.

lwshang (approved):

PR #1366, opened 2026-07-05
feat: wire the delegation permissions field through Delegation

aterga:
# Description

The `Delegation` class dropped the optional `permissions` field defined by the IC interface specification. As a result a **restricted (e.g. queries-only / read-only) delegation could not be represented, serialized, or forwarded** to a relying party by `@icp-sdk/core`.

This matters because the canister signature over a delegation **covers** the `permissions` field: the Internet Identity canister folds `permissions: "queries"` into the representation-independent hash it signs. When the agent reconstructs the delegation without that field, it recomputes a different hash, so the signature fails to verify and the delegation is unusable (read-only delegations "fail closed" on the dapp side).

This PR threads `permissions` through the delegation types so the field round-trips end-to-end:

- **`Delegation`** — new optional constructor arg and readonly `permissions` property.
- **`toCborValue`** — includes `permissions` when present, so it appears on the wire in `sender_delegation` and the replica recomputes the signed hash correctly.
- **`toJSON` / `fromJSON`** and the `JsonnableDelegation` / `JsonnableDelegationChain` types — persists and restores the field. It is kept as a plain string (e.g. `"queries"`) rather than hex-encoded, because it is a semantic value, not opaque binary; `fromJSON` validates it is a string.
- **`DelegationChain.create`** — new `permissions` option, for symmetry when minting chains.

All additions are optional and omitted when absent, so existing (unrestricted) delegations serialize **byte-for-byte** as before — no behavioral change for current callers.

### Context

Discovered while investigating why Internet Identity read-only delegations were not honored end-to-end: II already signs and requests `permissions`, but the agent library had nowhere to carry it. This is the library-side prerequisite noted in the II frontend (`@icp-sdk/core` must round-trip the `permissions` field before restricted delegations can verify on the relying-party side).

Fixes # (no issue filed yet)

# How Has This Been Tested?

Added unit tests in `packages/core/src/identity/identity/delegation.test.ts`:

- `permissions` is exposed on the `Delegation` instance.
- Included in `toCborValue` / `toJSON` when present; **omitted when absent** (backwards-compatibility assertions).
- Round-trips through `DelegationChain.create` → `toJSON` → `fromJSON`.
- `fromJSON` throws `Invalid permissions.` on a non-string value.
- **Cross-implementation reference vector**: `requestIdOf({...delegation})` for a queries-only delegation reproduces the exact representation-independent hash (`061d792a…33a4c6`) pinned by the Internet Identity backend test `should_pin_queries_permissions_msg`, guaranteeing the agent recomputes byte-for-byte what the canister signs.

Commands run in `packages/core`:

- `pnpm typecheck` — clean
- `pnpm test` — **42 suites / 638 tests pass** (incl. the pre-existing delegation golden test, confirming no serialization drift)
- `eslint` + `prettier --check` on the changed files — clean

# Checklist:

- [ ] I have edited the CHANGELOG accordingly. _(The changelog is auto-generated from conventional commits per CONTRIBUTING.md — no manual edit.)_
- [ ] I have made corresponding changes to the documentation. _(API docs are generated from the source TSDoc, which is updated here.)_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01YRakSoztjTRdffETFZXGjL)_

### DIFF
--- a/packages/core/src/identity/identity/delegation.ts
+++ b/packages/core/src/identity/identity/delegation.ts
@@ -58,7 +58,7 @@ export class Delegation implements ToCborValue {
       ...(this.targets && {
         targets: this.targets,
       }),
-      ...(this.permissions !== undefined && {
+      ...(this.permissions && {
         permissions: this.permissions,
       }),
     };
@@ -74,7 +74,7 @@ export class Delegation implements ToCborValue {
       expiration: this.expiration.toString(16),
       pubkey: safeBytesToHex(this.pubkey),
       ...(this.targets && { targets: this.targets.map(p => p.toHex()) }),
-      ...(this.permissions !== undefined && { permissions: this.permissions }),
+      ...(this.permissions && { permissions: this.permissions }),
     };
   }
 }
@@ -287,7 +287,7 @@ export class DelegationChain {
             ...(targets && {
               targets: targets.map(t => t.toHex()),
             }),
-            ...(permissions !== undefined && { permissions }),
+            ...(permissions && { permissions }),
           },
           signature: safeBytesToHex(signature),
         };
--- a/packages/core/src/identity/identity/delegation.ts
+++ b/packages/core/src/identity/identity/delegation.ts
@@ -46,6 +46,7 @@ export class Delegation implements ToCborValue {
     pubkey: Uint8Array,
     public readonly expiration: bigint,
     public readonly targets?: Principal[],
+    public readonly permissions?: string,
   ) {
     this.pubkey = pubkey.slice();
   }
@@ -57,6 +58,9 @@ export class Delegation implements ToCborValue {
       ...(this.targets && {
         targets: this.targets,
       }),
+      ...(this.permissions !== undefined && {
+        permissions: this.permissions,
+      }),
     };
   }
 
@@ -64,10 +68,13 @@ export class Delegation implements ToCborValue {
     // every string should be hex and once-de-hexed,
     // discoverable what it is (e.g. de-hex to get JSON with a 'type' property, or de-hex to DER
     // with an OID). After de-hex, if it's not obvious what it is, it's an ArrayBuffer.
+    // `permissions` is the exception: it is a semantic value (e.g. "queries"),
+    // not opaque binary, so it is kept as a plain string rather than hex-encoded.
     return {
       expiration: this.expiration.toString(16),
       pubkey: safeBytesToHex(this.pubkey),
       ...(this.targets && { targets: this.targets.map(p => p.toHex()) }),
+      ...(this.permissions !== undefined && { permissions: this.permissions }),
     };
   }
 }
@@ -85,6 +92,9 @@ interface JsonnableDelegation {
   pubkey: string;
   // Array of strings, where each string is hex of principal blob (*NOT* textual representation).
   targets?: string[];
+  // The delegation's optional `permissions` field (e.g. "queries"). Unlike the
+  // fields above, this is a plain semantic string, *NOT* hex-encoded.
+  permissions?: string;
 }
 
 /**
@@ -104,17 +114,20 @@ export interface SignedDelegation {
  * @param to The identity that receives the delegation.
  * @param expiration An expiration date for this delegation.
  * @param targets Limit this delegation to the target principals.
+ * @param permissions Restrict the kinds of calls the delegation permits (e.g. "queries").
  */
 async function _createSingleDelegation(
   from: SignIdentity,
   to: PublicKey,
   expiration: Date,
   targets?: Principal[],
+  permissions?: string,
 ): Promise<SignedDelegation> {
   const delegation: Delegation = new Delegation(
     to.toDer(),
     BigInt(+expiration) * BigInt(1000000), // In nanoseconds.
     targets,
+    permissions,
   );
   // The signature is calculated by signing the concatenation of the domain separator
   // and the message.
@@ -141,6 +154,7 @@ export interface JsonnableDelegationChain {
       pubkey: string;
       expiration: string;
       targets?: string[];
+      permissions?: string;
     };
   }>;
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/identity/identity/delegation.ts
```diff
@@ -57,17 +58,23 @@ export class Delegation implements ToCborValue {
...
     // every string should be hex and once-de-hexed,
     // discoverable what it is (e.g. de-hex to get JSON with a 'type' property, or de-hex to DER
     // with an OID). After de-hex, if it's not obvious what it is, it's an ArrayBuffer.
+    // `permissions` is the exception: it is a semantic value (e.g. "queries"),
+    // not opaque binary, so it is kept as a plain string rather than hex-encoded.
     return {
       expiration: this.expiration.toString(16),
       pubkey: safeBytesToHex(this.pubkey),
       ...(this.targets && { targets: this.targets.map(p => p.toHex()) }),
+      ...(this.permissions !== undefined && { permissions: this.permissions }),
```
aterga: Why don't we follow the local convention here?


```suggestion
      ...(this.permissions && { permissions: this.permissions }),
```

packages/core/src/identity/identity/delegation.ts
```diff
@@ -253,14 +279,15 @@ export class DelegationChain {
...
-        const { targets } = delegation;
+        const { targets, permissions } = delegation;
         return {
           delegation: {
             expiration: delegation.expiration.toString(16),
             pubkey: safeBytesToHex(delegation.pubkey),
             ...(targets && {
               targets: targets.map(t => t.toHex()),
             }),
+            ...(permissions !== undefined && { permissions }),
```
aterga: Ditto 


```suggestion
            ...(permissions && { permissions }),
```

packages/core/src/identity/identity/delegation.ts
```diff
@@ -208,10 +230,13 @@ export class DelegationChain {
 
     const parsedDelegations: SignedDelegation[] = delegations.map(signedDelegation => {
       const { delegation, signature } = signedDelegation;
-      const { pubkey, expiration, targets } = delegation;
+      const { pubkey, expiration, targets, permissions } = delegation;
       if (targets !== undefined && !Array.isArray(targets)) {
         throw new Error('Invalid targets.');
       }
+      if (permissions !== undefined && typeof permissions !== 'string') {
```
aterga, line 237: permissions are a variant in terms of Candid, not string. 

please double check that this is the correct representation in JS

packages/core/src/identity/identity/delegation.ts
aterga: Done in 305332b — applied here and to the `toCborValue` and `DelegationChain.toJSON` spreads as well. Kept the explicit `!== undefined` only for the `fromJSON` validation guard, mirroring how `targets` is handled.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/identity/identity/delegation.ts
aterga: Done in 305332b (same change).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/identity/identity/delegation.ts
aterga: Double-checked — these are two different types, and the `Delegation` class needs the string form:

- `Permissions = variant { queries; all }` is only the **request-time method argument** (`prepare_account_delegation(..., permissions : opt Permissions)`) — how the caller *asks* II for a restricted delegation.
- The **issued delegation** carries a different field. From `internet_identity.did`:
  ```candid
  type Delegation = record {
    pubkey : PublicKey;
    expiration : Timestamp;
    targets : opt vec principal;
    permissions : opt text;   // "queries" when present; absent = unrestricted
  };
  ```
  So on the delegation object it's `opt text`, not the variant. (II maps the `all` variant to *no* `permissions` field, and `queries` to `permissions = "queries"`.)

This `Delegation` class represents that issued object — it's what goes into `sender_delegation` and whose representation-independent hash the II canister signs. It isn't Candid-encoded on the wire; it's CBOR plus the IC representation-independent hash, where the value is hashed as a plain string (`Value::String("queries")` on the II side). So `string` is the correct JS representation, and `opt text` → `string | undefined` maps cleanly.

The added test pins exactly this: `requestIdOf({ pubkey, expiration, permissions: "queries" })` reproduces the II backend's signed hash byte-for-byte (`061d792a…33a4c6` — the same vector as II's `should_pin_queries_permissions_msg`). If the representation were wrong the hash wouldn't match and the canister signature would fail to verify.

I used `string` rather than a `'queries'` literal so the type tracks the spec's `opt text` and won't need widening if more permission values are defined later. Happy to narrow it to a union if you'd prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

marc0olo (approved):
Approving. This is a clean, well-scoped change and the pinned-hash test against `should_pin_queries_permissions_msg` is exactly the right way to lock the cross-system contract — it proves the agent reconstructs byte-for-byte what the II canister signs, which is the only thing that could have gone subtly wrong here.

Verified on my end:
- Purely additive/optional — existing unrestricted delegations serialize byte-for-byte unchanged, so no risk to current callers.
- Signing hash (`{ ...delegation }` → `hashOfMap`, which filters `undefined`) stays consistent with the omit-when-absent wire encoding in `toCborValue`/`toJSON`.
- No other consumer of the `Delegation` shape breaks on the appended optional arg.

One optional nit, take it or leave it: the spread guards in `toCborValue`/`toJSON` are truthy checks (`this.permissions && …`), while `hashOfMap` only filters `undefined`. A degenerate `permissions: ''` would therefore be signed-over but omitted from the wire, yielding a signature that fails to verify. No real caller passes an empty string, so this isn't a blocker — but switching the two guards to `!== undefined` would make the omit/hash semantics identical and close the gap. Happy to merge with or without it.

Thanks for threading this through end-to-end.

PR #1367, opened 2026-07-06
ci: add weekly and manual npm audit with auto-fix PRs

lwshang:
# Description

Reworks the **NPM Audit** workflow so it does two jobs well: gate PRs that actually change dependencies, and keep the tree clean over time — without failing unrelated feature PRs (e.g. #1366).

Note: the NPM Audit check is **not** a required check on `main` (required = `unit` / `e2e` / `format` / `mitm`), so it has never gated merges — it is informational. This change targets that information where it's useful and removes it where it was just noise.

### Triggers

- **`pull_request`, scoped to dependency changes** (`paths: ['**/package.json', 'pnpm-lock.yaml']`). Catches vulnerabilities a PR *introduces*, right where the author can fix them. Code-only PRs no longer trigger the audit, so they're never flagged for newly-published advisories unrelated to the change — which was the root cause of the red X on #1366.
- **`schedule` (weekly, Mon 09:00 UTC)** — sweeps for advisories published against already-installed dependencies, independent of PR activity.
- **`workflow_dispatch`** — on-demand remediation.

### Auto-remediation (scheduled / manual only)

When a scheduled or manual run finds vulnerabilities, it runs `pnpm audit --fix`, refreshes the lockfile with `pnpm install`, and opens a bot PR via `dfinity/ci-tools/actions/create-pr` (the same `PR_AUTOMATION_BOT_PUBLIC` app used by [create-release-pr.yml](.github/workflows/create-release-pr.yml) and [generate-changelog.yml](.github/workflows/generate-changelog.yml)). You review that PR — with normal CI running on it — and merge to restore a clean tree.

If `pnpm audit --fix` produces no changes, the run **fails visibly** instead of opening an empty PR, signalling that manual intervention is needed.

The `pull_request` path never opens a bot PR — it only runs the pass/fail audit, preserving today's behavior for dependency-changing PRs.

## How Has This Been Tested?

- YAML validated locally; triggers confirmed (`pull_request` paths-scoped, `schedule`, `workflow_dispatch`).
- The remediation path can be exercised after merge via **Run workflow** (`workflow_dispatch`) — the current tree has open advisories, so the first manual run should produce the first `pnpm audit --fix` PR.

## Checklist:

- [ ] I have edited the CHANGELOG accordingly. _(CI-only change; changelog is auto-generated from conventional commits.)_
- [ ] I have made corresponding changes to the documentation. _(No user-facing docs affected.)_

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marc0olo (approved):
Approving — the PR-vs-sweep split, the paths-scoping, and failing visibly on a no-op `--fix` (instead of opening an empty PR) are all the right calls. A couple of notes to verify, none blocking:

**1. The remediation job likely needs `environment: Release` to read the bot key.** `PR_AUTOMATION_BOT_PUBLIC_PRIVATE_KEY` appears to be a **`Release` environment secret**, not a repo-level secret — and the `audit` job declares no environment. If that's the case, the secret resolves empty on scheduled/manual runs, `create-github-app-token` fails, and no fix PR is opened. Every other workflow using this key (`create-release-pr`, `generate-changelog`, `publish-docs`) sets `environment: Release`, which is what points to this.

Cleanest fix: split the token + create-PR steps into a separate job with `needs: audit`, gated on the non-PR + `needs_update == 'true'` condition, and `environment: Release` — that keeps the PR-gating audit environment-free while giving the auto-fix job (which only runs on `main`) access to the key. Adding `environment: Release` to the single job also works. Worth confirming against the repo's secret scoping.

**2. Confirm the GitHub App has `pull-requests: write` (plus `contents: write`).** The create-PR step runs on the app token, so the App's permissions govern branch push + PR creation, not the job's `permissions:` block. `create-release-pr` already opens PRs with this app, so this is probably fine.

**3. Treat the first `workflow_dispatch` run as the acceptance test.** Since the tree currently has open advisories, a manual run should produce the first real fix PR — that green run is the actual proof the remediation path works end-to-end.

None of these affect PR gating or `main`, so fine to land and validate via the manual dispatch. Thanks for cleaning this up!

PR #1375, opened 2026-07-14
feat: Revamp Agent.readState

adamspofford-dfinity:
The previous API, `canisterStatus.request`:
* was a free function, instead of an agent method
* took magic strings, even though paths are also strings, and the relationship was confusing
* doesn't have support for anything parameterized
* is unintuitively mandatory because it checks certificates where `readState` does not check certificates
* is named wrong

This PR expands `Agent.readState` to a second attempt at the same style of API. Parsing objects are used instead of magic strings, allowing e.g. specific named canister metadata to be requested. It also now checks certificates. The old API is marked `@deprecated` and reimplemented in terms of the new one.


packages/core/src/agent/canisterStatus/index.ts
```diff
@@ -39,6 +36,42 @@ export type Path = 'time' | 'controllers' | 'subnet' | 'module_hash' | 'candid'
...
+  path: Exclude<Path, CustomPath | 'subnet'>,
+  canisterId: Principal,
+): KnownPath<Status> => {
+  switch (path) {
+    case 'time':
+      return StatePaths.time;
+    case 'controllers':
+      return StatePaths.canisterControllers(canisterId);
+    case 'module_hash':
+      return StatePaths.canisterModuleHash(canisterId);
```
lwshang: **Likely unintended breaking change: `module_hash` return type changes from hex `string` to `Uint8Array`.**

On `main`, `canisterStatus.request(['module_hash'])` returns `bytesToHex(data)` — a hex string. Here `namedToKnown` maps `module_hash` to `StatePaths.canisterModuleHash(canisterId)`, whose decoder is `x => x` (in `utils/readState.ts`), so this deprecated-but-still-public API (re-exported from the package root) now returns a raw `Uint8Array`.

The other named paths (`time`, `controllers`, `candid`, and custom paths) all preserve their previous outputs, so this one looks accidental — it seems to fall out of reusing `StatePaths.canisterModuleHash`, which intentionally returns bytes for the *new* API. It's also no longer guarded by a test: the behavioral `should query canister module hash` snapshot that pinned the hex string was deleted in this PR.

Suggest preserving the hex-string output for the deprecated function, e.g.:

```ts
case 'module_hash':
  return new KnownPath(
    ['canister', canisterId.toUint8Array(), 'module_hash'],
    bytesToHex,
  );
```

or, if the change is intentional, please call it out as a breaking change in the changeset.

lwshang (changes requested):
Requesting changes on the `module_hash` return-type regression flagged inline — it changes a public (if deprecated) API from a hex string to a `Uint8Array` and is no longer test-covered.

A few other things to address or at least document before merge — fine as follow-ups if you prefer, but calling them out:

- **Missing end-to-end coverage for the `canisterRanges` path fix (subnetStatus).** The encoded path was corrected from `[canister_ranges, subnetId]` to `[subnet, subnetId, canister_ranges]` (this matches the IC spec, so it's a genuine latent-bug fix). But the only coverage is the self-consistent `encodePath` unit test plus a standalone `decodeCanisterRanges` test that decodes raw bytes directly. There's no `subnetStatus.request(['canisterRanges'])` test against a golden certificate confirming the corrected path actually resolves a value end-to-end.

- **`Agent.readState` contract change is breaking for custom agents.** `pollForResponse` now reads `state.verifiedCertificate` directly and drops its own `Certificate.create` + root-key guard. Any third-party `Agent` implementation that returns only `{ certificate }` (valid under the old interface) will make polling throw on `undefined.lookup_path`. This is a reasonable consequence of the revamp, but please surface it in the migration/release notes so agent implementers know they must populate `verifiedCertificate`.

- **`readState` now always verifies (BLS + freshness) without first triggering time sync.** Direct callers now pay for verification they previously didn't, and a first-operation `readState` with a skewed local clock will fail the freshness check (it calls `#rootKeyGuard()` but not `#syncTimeGuard()`) rather than returning bytes. Internal bootstrapping paths correctly pass `disableCertificateTimeVerification: true`, so this only affects external direct callers — worth documenting alongside the new `disableTimeVerification` option.

adamspofford-dfinity:
Changelog info goes in the autogenerated pr later, I don't know if there's a way to mark it as to-be-added from here.

lwshang (approved):

